PR #4, opened 2017-10-13
SF-1321: implement changing of passwords in the cli

rklind:


src/commands/users/password.ts
```diff
@@ -0,0 +1,7 @@
+export = {
+	command: 'password',
+	desc: 'Manage SkySync Users\' Passwords',
+	builder: yargs => yargs.commandDir('./password', {
+		exclude: /util/
+	}).demandCommand(1)
+}
```
rklind: @tclev thank you for solving this riddle!

PR #5, opened 2017-10-23
Added SDK sample

avolinsky:


samples/node/createJob.js
```diff
@@ -0,0 +1,61 @@
+const credentials = require('./credentials').credentials;
+const SkySyncClient = require('skysync-cli').SkySyncClient;
+
+const client = new SkySyncClient(credentials);
+
+var nfsConnection;
```
smbecker: Use `let` rather than `var`

samples/node/createJob.js
```diff
@@ -0,0 +1,61 @@
...
+var o365Connection;
+
+var nfsConnectionPromise = client.connections.add({
+	name: 'Sample NFS Connection',
+	platform: { id: 'nfs' },
+	auth: {
+		uri: '//test-server',
```
smbecker: NFS uses back slash rather than forward slash

samples/node/credentials.js
```diff
@@ -0,0 +1,6 @@
+const credentials = {
```
smbecker: Why not just `export = {}` so that you can do `const credentials = require('./credentials');`?

samples/node/deleteSampleData.js
```diff
@@ -0,0 +1,24 @@
...
+		}, this);
+	})
+	.catch(console.error);
+
+client.jobs.list({active: true, kind: 'transfer'})
+	.then(jobs => {
+		jobs.forEach(function(job) {
+			if (job.name.startsWith('Sample')) {
+				console.log(`Deleting job ${job.name} (${job.id})`);
+				client.jobs.delete(job.id);
```
smbecker: Should we be including catches on these?

samples/node/createJob.js
```diff
@@ -0,0 +1,61 @@
...
+		username: 'localhost/admin',
+		password: 'secret'
+	}
+}).then(connection => { nfsConnection = connection });
```
mikesae: So I'm wondering why we don't use a script like [this](https://github.com/portalarchitects/usfs/blob/master/scripts/create-connections.js) where for example you'd create connections with the skysync-cli command like

```javascript
function createConnection(platform, name, authFile) {
	let cmd = `skysync-cli connections add --platform ${platform } --name ${name} --auth-file ${authFile}`;
	console.log(cmd);
	return exec(cmd);
}
```

smbecker: This is showing how you could create jobs using the SDK rather than the CLI.

PR #7, opened 2017-11-09
SF-1332: New CLI features to manage connection pools

jtirjan:


src/sdk/models/connections.ts
```diff
@@ -15,6 +15,7 @@ export interface Connection extends IEntityIdentifier<string> {
 	features?: ConnectionFeatures;
 	platform?: StoragePlatform;
 	account?: Account;
+	poolId?: string;
```
smbecker: This isn't right

smbecker: Should be pool object

src/commands/connections/pool/set.ts
```diff
@@ -0,0 +1,19 @@
+import { runCommand } from '../../../util/command';
+import { outputFormat } from '../util';
+const open = require('open')
+
+export = {
+	command: 'set <id> <poolId>',
```
smbecker: Allow setting pool name as well

src/commands/connections/util.ts
```diff
@@ -0,0 +1,32 @@
...
+			property: 'disabled',
+			transform: val => !val
+		},
+		{
+			header: 'IsPool',
+			property: 'group'
+		},
+		{
+			header: 'ParentPool',
+			property: 'pool.id'
```
smbecker: I would show pool name here and pool id in the json below

smbecker: Or at least show pool name in the json

jtirjan: Made the change, but it appears the REST service is returning connection.Name in the connection.Pool.Name field. Looking into that now...

src/commands/connections/pool/clear.ts
```diff
@@ -0,0 +1,18 @@
+import { runCommand } from '../../../util/command';
+import { outputFormat } from '../util';
+const open = require('open')
+
+export = {
+	command: 'clear <id>',
```
rklind: i propose changing this to either delete or remove vs. clear. delete is more consistent throughout the application, but may not exactly reflect the action being taken...

src/commands/connections/pool/set.ts
```diff
@@ -0,0 +1,21 @@
+import { runCommand } from '../../../util/command';
+import { outputFormat } from '../util';
+const open = require('open')
+
+export = {
+	command: 'set <id> [poolId] [poolName]',
```
rklind: i suggest using add or assign instead of set.  also, i would think you'd only want to add by specifying the id of the pool instead of the name, here.  can a connection be  a part of more than one pool?

src/commands/connections/util.ts
```diff
@@ -0,0 +1,33 @@
...
+			header: 'Enabled',
+			property: 'disabled',
+			transform: val => !val
+		},
+		{
+			header: 'IsPool',
+			property: 'group'
+		},
+		{
+			header: 'ParentPool',
```
rklind: maybe change this heading to PoolName?

src/commands/connections/util.ts
```diff
@@ -0,0 +1,33 @@
...
+			property: 'pool.name',
```
rklind: can a connection be  a part of more than one pool?

src/commands/connections/pool/clear.ts
jtirjan: Remove sounds better to me. Delete gives a stronger implication that it's deleting the actual connection pool, i.e.: "skysync-cli connections pool delete 2345432632"

src/commands/connections/util.ts
jtirjan: No, a connection can only be part of 1 pool

src/commands/connections/pool/set.ts
jtirjan: if you specify by name only, it will create a new pool with that name (if one doesn't exist). i believe that's the only way to create a new pool right now

src/commands/connections/util.ts
jtirjan: Data reference between connection & pool was using the wrong column. Fixed server-side, no client changes needed

src/commands/connections/pool/set.ts
smbecker: I like `assign` and `unassign` since we are assigning a connection to a new/existing pool

smbecker: I would change the `[poolId] [poolName]` options into explicit options so you can put in descriptions for them. I would also recommend changing the option names to `pool: existing pool ID` and `name: new pool name`

src/commands/connections/pool/set.ts
```diff
@@ -0,0 +1,21 @@
...
+	desc: 'Assign a connection to a connection pool',
+	handler: argv => {
+		runCommand(argv, async (client, output) => {
+			const connection = await client.connections.setPool(argv.id,
+				{
+					id: argv.poolId,
+					name: argv.poolName
+				},
+				{
+					fields: ['name']
```
smbecker: Do we need more `fields` to satisfy the `outputFormat`?

src/commands/connections/pool/assign.ts
```diff
@@ -0,0 +1,58 @@
...
+			header: 'ConnectionName',
+			property: 'name'
+		},
+		{
+			header: 'Enabled',
+			property: 'disabled',
+			transform: val => !val
+		},
+		{
+			header: 'PoolId',
```
smbecker: Convention is to use `ID` rather than `Id`.

src/commands/connections/pool/assign.ts
```diff
@@ -0,0 +1,58 @@
...
+		}
+	]
+};
+
+export = {
+	command: 'assign <id>',
+	desc: 'Assign a connection to an existing connection pool, or create one if it does not exist.',
+	builder: yargs => {
+		yargs.options({
+			'poolId': {
```
smbecker: I would use `pool` and `name` for these arguments.

src/sdk/models/connections.ts
```diff
@@ -15,12 +15,18 @@ export interface Connection extends IEntityIdentifier<string> {
...
 	account?: Account;
+	pool?: ConnectionPool;
 }
 
 export interface Account extends IEntityIdentifier<string> {
 	email?: string;
 }
 
+export interface ConnectionPool extends IEntityIdentifier<string> {
+	id?: string;
```
smbecker: This property shouldn't be required since it is inherited from `IEntityIdentifier<string>`.

PR #8, opened 2017-11-15
SF-1323:  CLI Feature to allow starting and pausing multiple jobs by filter

jtirjan:


src/commands/jobs/pause.ts
```diff
@@ -1,19 +1,67 @@
...
+	desc: 'Pause job(s)',
+	builder: yargs => {
+		yargs.options({
+			'all': {
+				desc: 'Pause all jobs',
+				type: 'boolean',
+				group: 'Search',
+				default: undefined
+			},
+			'parent': {
```
smbecker: I would just add `parent` to job search args.

src/commands/jobs/pause.ts
```diff
@@ -1,19 +1,67 @@
...
-			const job = await client.jobs.pause(argv.id, {
-				fields: [
-					'name',
-					'status',
-					'disabled'
-				]
+		if (argv.id == undefined) {
+			if (searchCriteriaIsMissing(argv)) {
+				console.error("Search criteria must be specified. To pause all jobs, use --all parameter");
+			}else{
```
smbecker: Missing spaces between braces

src/commands/jobs/pause.ts
```diff
@@ -1,19 +1,67 @@
...
+import { detailOutputFormat, jobsSearchArgumentsDefault, searchCriteriaIsMissing } from './util';
 
 export = {
-	command: 'pause <id>',
-	desc: 'Pause job',
+	command: 'pause [id]',
+	desc: 'Pause job(s)',
+	builder: yargs => {
+		yargs.options({
+			'all': {
```
smbecker: I don't think this is a valid parameter.

src/commands/jobs/util.ts
```diff
@@ -138,5 +155,13 @@ const auditOutputFormat = {
 	]
 };
 
-export { listOutputFormat, detailOutputFormat, historyOutputFormat, auditOutputFormat };
+function searchCriteriaIsMissing(argv):boolean {
+	return	argv.all == undefined
+		&& argv.parent == undefined
```
smbecker: Use `===` rather than `==`

src/commands/jobs/pause.ts
jtirjan: @rklind and I were talking about forcing the user to specify --all to start/pause all jobs as a safety precaution. Should I remove that safety check, or maybe implement it a different way?

smbecker: That's fine but you pass it down to the parameters block that is serialized to the service. We shouldn't do that.

jtirjan: Ah, gotcha

smbecker: What if rather than using `--all` option, it keyed off of the `id` positional argument? That would allow you to keep `pause <id>` so that the positional argument is required so that you don't have to check if it is available below. You would simply check `argv.id === 'all'`. This would change the usage from `skysync-cli start --all` to `skysync-cli start all`. Thoughts?

jtirjan: It's possible to call start/pause without `id` or `all`.
For example, `skysync-cli jobs start --parent=1234`
I'm leaning towards leaving the `--all` switch as-is, but I'm open to discussing this further

src/commands/jobs/pause.ts
```diff
@@ -1,19 +1,59 @@
...
-				fields: [
-					'name',
-					'status',
-					'disabled'
-				]
+		if (argv.id === undefined) {
+			if (searchCriteriaIsMissing(argv)) {
+				console.error("Search criteria must be specified. To pause all jobs, use --all parameter");
+			}
+			else {
```
smbecker: Formatting... `else` should be on the same line as the bracket.

src/commands/jobs/pause.ts
```diff
@@ -1,19 +1,59 @@
...
-		runCommand(argv, async (client, output) => {
-			const job = await client.jobs.pause(argv.id, {
-				fields: [
-					'name',
-					'status',
-					'disabled'
-				]
+		if (argv.id === undefined) {
+			if (searchCriteriaIsMissing(argv)) {
+				console.error("Search criteria must be specified. To pause all jobs, use --all parameter");
```
smbecker: I would output this as red text similar to what we do in `writeFailure` [here](https://github.com/portalarchitects/skysync-cli/blob/master/src/util/formatter.ts#L47)

src/commands/jobs/pause.ts
```diff
@@ -1,19 +1,59 @@
...
+			else {
+				runCommand(argv, async (client, output) => {
+					const result = await client.jobs.pauseMultiple({
+						...getSearchArgs(argv)
+					});
+					const totalCount = result.meta.total_count;
+					if (totalCount === undefined || totalCount == 0) {
+						output.writeWarning("Warning: 0 jobs were paused because no jobs matched the specified filter");
+					}
+					else if (totalCount == 1) {
```
smbecker: `else` should be on the same line as bracket

PR #10, opened 2017-11-17
SF-1548 - Added 'engines' restriction and integrated lint command

smbecker:


package.json
```diff
@@ -32,12 +32,17 @@
     "dist/",
     "bin/"
   ],
+  "engineStrict": true,
+  "engines": {
+    "node": ">=8.3.0"
+  },
```
rklind: I thought this was all that was needed.  I didn't realize something needed to be added to skysync.js.  :(

smbecker: Apparently, it is only a suggestion and not always enforced anyway.

http://www.marcusoft.net/2015/03/packagejson-and-engines-and-enginestrict.html

PR #12, opened 2017-11-22
CLI Enhancement to allow managing Audit Categories

jtirjan:


jtirjan:
Looking into the Travis errors...

src/sdk/client.ts
```diff
@@ -70,4 +71,8 @@ export class SkySyncClient {
 	get users(): UsersResource {
 		return new UsersResource(this._httpClient);
 	}
+
+	get categories(): CategoriesResource {
```
smbecker: `auditCategories`

src/sdk/models/categories.ts
```diff
@@ -0,0 +1,5 @@
+import { IEntityIdentifier } from './base';
+
+export interface Category extends IEntityIdentifier<string> {
```
smbecker: `Category` is too generic since we also have `JobCategory` and potentially others down the road. I would change the class and resource names to be prefixed with `Audit`. So, `AuditCategory` and `AuditCategoriesResource`. I would also change the command to `audit-categories`

PR #13, opened 2017-11-30
SF-1328 - CLI: Implement an interface to the metrics endpoint

recordrn:
Followed the pattern similar to the Connections Platform command regarding file structure and naming conventions.


src/commands/diagnostics/metrics.ts
```diff
@@ -0,0 +1,67 @@
...
+			header: 'Type',
+			property: 'type'
+		}
+	],
+	json: [
+	]
+};
+
+export = {
+	command: 'metrics',
```
jtirjan: For consistency, I'd suggest a command structure such as `skysync-cli diagnostic-metrics list`

src/sdk/models/diagnostics.ts
```diff
@@ -0,0 +1,5 @@
+import { IEntityIdentifier } from './base';
+
+export interface DiagnosticMetric extends IEntityIdentifier<string> {
+	metrics: any;
```
jtirjan: I expected fields like name, value, etc here.
Also, is `IEntityIdentifier` inheritance needed? I don't recall seeing an `id` in the data.

src/commands/diagnostics/metrics.ts
recordrn: I followed a similar pattern to the Connection Platforms since metrics is similar in relationship to diagnostics.

src/sdk/models/diagnostics.ts
recordrn: good point. I will look into the IEntityIdentifier and interface for the model.

src/commands/diagnostics/metrics.ts
recordrn: The full command is: 'skysync-cli diagnostics metrics'. There are other diagnostic endpoints that will be commands under the diagnostics namespace/container.

src/commands/diagnostics/metrics.ts
```diff
@@ -0,0 +1,67 @@
...
+		{
+			header: 'Value',
+			property: 'value'
+		},
+		{
+			header: 'Type',
+			property: 'type'
+		}
+	],
+	json: [
```
smbecker: Is this property needed if it is empty?

src/commands/diagnostics/metrics.ts
```diff
@@ -0,0 +1,67 @@
...
+				name: argv.name,
+				q: argv.search
+			});
+			
+			let mappedMetrics =
+				Object.keys(metrics).map(name => {
+					const metric = metrics[name];
+					return {
+						name: name,
+						value: metrics[name].value,
```
smbecker: You don't use the `metric` variable. Can you just do `...metric`?

src/commands/diagnostics/metrics.ts
```diff
@@ -0,0 +1,67 @@
...
+						type: metrics[name].type
+					};
+				});
+			output.writeItem(mappedMetrics, outputFormat);
```
smbecker: `writeTable`

src/sdk/resources/diagnostics.ts
```diff
@@ -0,0 +1,11 @@
+import { IHttpClient } from '../http';
+import { EditableResource } from './resource';
+import { DiagnosticMetric } from '../models';
+
+export class DiagnosticMetricsResource extends EditableResource<DiagnosticMetric> {
```
smbecker: Metrics aren't editable. You should derive from `Resource` rather than `EditableResource`

smbecker: Actually, after looking at `Resource`, I'm not sure that is the right class either because it exposes a `get(id)` method, which isn't valid for metrics. I would actually refactor `Resource` into a `BaseResource` class that looks like this:

```javascript
export class BaseResource {
	public defaultParams: any;

	constructor(protected httpClient: IHttpClient) {}

	protected mergeDefaultParams(params: any): any {
		return this.mergeParams(this.defaultParams, params);
	}

	protected mergeParams(lhs: any, rhs: any): any {
		if (!lhs) {
			return rhs;
		}

		if (!rhs) {
			return lhs;
		}

		return Object.assign({}, lhs, rhs);
	}
}
```

and then this class becomes

```javascript
export class DiagnosticMetricsResource extends BaseResource {
		super(httpClient);
	}

	async list(params?: any): Promise<TResource[]> {
		const result = await this.httpClient.get('diagnostics/metrics', this.mergeDefaultParams(params));
		return getTypedResponse<TResource[]>(result, 'metrics');
	}
}
```

src/commands/diagnostics/metrics.ts
jtirjan: Cool, that makes sense

src/commands/diagnostics.ts
```diff
@@ -0,0 +1,5 @@
+export = {
+	command: 'diagnostics',
+	desc: 'Manage SkySync Diagnostics',
```
rklind: the description should probably be something like View Skysync Diagnostics.  i don't see anything pertaining to editing / creating / deleting in this story.  we can always update this later when we add additional features.

recordrn: There are a few other diagnostic endpoints for deleting, updating log config, etc, but I'm fine with just changing to 'View' for now until these other commands are available in the cli. We will just have to remember to change it back.

src/commands/diagnostics/metrics.ts
recordrn: removed json property

src/commands/diagnostics/metrics.ts
recordrn: good catch. switched over to using the metric object.

src/commands/diagnostics/metrics.ts
recordrn: updated to .writeTable (although the output appears the same as writeItem with the defined outputFormat)

src/sdk/models/diagnostics.ts
recordrn: updated. removed inheritance from IEntityIdentifer and updated the model.

src/sdk/resources/diagnostics.ts
recordrn: BaseResource class added to the Resources.ts, and updated the resources/diagnostics.ts to the suggested approach(with a few small tweaks).

src/sdk/resources/resource.ts
```diff
@@ -8,6 +8,28 @@ export function getTypedResponse<T>(result: any, type?: string): T {
...
+		
+		if (!rhs) {
+			return lhs;
+		}
+		
+		return Object.assign({}, lhs, rhs);
+	}
+}
+
 export class Resource<TResource> {
```
smbecker: `Resource` should derive from `BaseResource` and remove the overlapping code.

src/sdk/resources/resource.ts
```diff
@@ -41,6 +28,25 @@ export class Resource<TResource> {
 
 		return Object.assign({}, lhs, rhs);
 	}
+}
+
+export class Resource<TResource> extends BaseResource {
+	constructor(protected httpClient: IHttpClient, protected singularName: string, protected pluralName: string = undefined,
```
smbecker: Does this work with `protected` qualifier on this field since we technically just want to use the base `httpClient`? Shouldn't this just be `httpClient: IHttpClient` so that it doesn't attempt to assign it to any local field since we already do that as a result of the call to `super(httpClient);`?

recordrn: It does work (I checked on a cmd that uses resource before I committed); however, I see why this makes sense since it inherits the BaseResource constructor. `protected` removed.

PR #14, opened 2017-11-30
SF-1694: CLI Enhancement to download Remote Site Packages

jtirjan:


src/commands/sites/download.ts
```diff
@@ -0,0 +1,61 @@
...
+					fs.mkdirSync(pathSegment);
+				} catch (err) {
+					if (err.code !== 'EEXIST') {
+						throw err;
+					}
+				}
+				return pathSegment;
+			}, path.isAbsolute(argv.targetDirectory) ? path.sep : '');
+
+			(await client.sites.getFile(argv.id))
```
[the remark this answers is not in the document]
smbecker: Perhaps add an interface like the following in `resource.ts`:

```javascript
export interface IDownloadFileProvider {
	getDownloadRequestPath(id: string): string;
}
```
Then add the following in `util/file-downloader.ts`

```javascript
const path = require('path');
const fs = require('fs');

export class FileDownloader {
	constructor(private httpClient: IHttpClient, private fileProvider: IDownloadFileProvider) {}

	download(id: string, outputDirectory: string): Promise<any> {
		const downloadPath = fileProvider.getDownloadRequestPath(id);
		return new Promise<any>(async (resolve, reject) => {
			try {
				await httpClient.download((fileName, output) => {
					const outputPath = path.join(outputDirectory, fileName);
					output.pipe(fs.createWriteStream(outputPath))
						.on('finish', resolve)
						.on('error', reject);
				}, downloadPath);
			} catch (err) {
				return reject(err);
			}
		});
	}
}
```

Then `SitesResource` would derive from `IDownloadFileProvider` with the following:

```javascript
getDownloadRequestPath(id: string): string => `${this.resourcePath}/${id}/download`;
```

src/sdk/resources/sites.ts
```diff
@@ -1,9 +1,20 @@
 import { IHttpClient } from '../http';
 import { EditableResource } from './resource';
 import { RemoteSite } from '../models';
+import { IDownloadFileProvider} from './resource';
```
smbecker: Merge this line with the `./resource` import above.

src/sdk/resources/sites.ts
```diff
@@ -1,9 +1,20 @@
...
+import { FileDownloader} from '../../util/file-downloader';
 
-export class SitesResource extends EditableResource<RemoteSite> {
+export class SitesResource extends EditableResource<RemoteSite> implements IDownloadFileProvider {
 	constructor(httpClient: IHttpClient) {
 		super(httpClient, 'site');
 	}
+
+	download(id: string, outputDirectory: string): any {
```
smbecker: I would remove this method and rely on callers to use the `FileDownloader` for themselves. The main reason is because if we are going to leverage the sdk pieces in the browser, file download would not be available in that case. So, I would like to keep the main file-download concerns/implementation out of the `sdk` directory.

src/util/file-downloader.ts
```diff
@@ -0,0 +1,69 @@
+import { IHttpClient } from '../sdk/http';
+import { IDownloadFileProvider } from '../sdk/resources';
+
+const path = require('path');
+const fs = require('fs');
+
+export class FileDownloader {
+	constructor(private httpClient: IHttpClient, private fileProvider: IDownloadFileProvider) {}
+
+	createDirectory(targetDirectory: string) {
```
smbecker: `private` or pull it up as a function outside of this class so that it doesn't get exported with the public contract.

src/util/file-downloader.ts
```diff
@@ -0,0 +1,69 @@
...
+			try {
+				let _this = this;
+
+				this.createDirectory(outputDirectory);
+				let requestPath = this.fileProvider.getDownloadRequestPath(id);
+				let response = await this.httpClient.getResponse(requestPath);
+
+				await response.on('response', async function (response) {
+					if (response.statusCode === 404) {
+						return reject(`Site with ID=${id} not found`);
```
smbecker: Use a more generic error message. `The requested file could not be found.`

src/util/file-downloader.ts
```diff
@@ -0,0 +1,69 @@
...
+					}
+					let fileName = _this.parseFileName(response);
+					if (!fileName) {
+						return reject('Server did not return a file');
```
smbecker: `The server...`

src/util/file-downloader.ts
```diff
@@ -0,0 +1,69 @@
...
+					}
+					const outputPath = path.join(outputDirectory, fileName);
+					await response.pipe(fs.createWriteStream(outputPath))
+						.on('finish', () => {
+							resolve(outputPath)
+						})
+						.on('error', (err) => {
```
smbecker: `.on('error', reject)`. The lambda shouldn't be necessary

src/util/file-downloader.ts
```diff
@@ -0,0 +1,69 @@
...
+							reject(err)
+						});
+				});
+			} catch (e) {
+				reject(e.message);
+			}
+		});
+	}
+
+	parseFileName(response: any): string {
```
smbecker: `private` or pull it up as a function outside of this class so that it doesn't get exported with the public contract.

src/sdk/http/http-client.ts
```diff
@@ -156,7 +158,29 @@ export abstract class HttpClient<TRequest, TResponse> implements IHttpClient {
 			}
 		});
 	}
-
+	
+	getResponse(path: string, params?: any): Promise<any> {
```
smbecker: Why still returning the raw `response` rather than the abstract method I proposed earlier? Was there a reason that wouldn't have worked? This still requires some internal implementation knowledge for consumers.

src/sdk/resources/sites.ts
```diff
@@ -1,9 +1,17 @@
...
+import { EditableResource, IDownloadFileProvider } from './resource';
 import { RemoteSite } from '../models';
 
-export class SitesResource extends EditableResource<RemoteSite> {
+export class SitesResource extends EditableResource<RemoteSite> implements IDownloadFileProvider {
 	constructor(httpClient: IHttpClient) {
 		super(httpClient, 'site');
 	}
+
+	getHttpClient(): any {
```
smbecker: This doesn't need to be exposed here. The `SkySyncClient` already exposes the http client.

src/util/file-downloader.ts
```diff
@@ -0,0 +1,66 @@
...
+
+	download(id: string, outputDirectory: string): Promise<string> {
+		return new Promise<any>(async (resolve, reject) => {
+			try {
+				let _this = this;
+				this.createDirectory(outputDirectory);
+				let requestPath = this.fileProvider.getDownloadRequestPath(id);
+
+				return await this.httpClient.download(requestPath, async (response) => {
+					if (response.statusCode === 404) {
```
smbecker: This logic should be handled in the http client. We do not want the consumer of `download` to be aware of the implementation details of the response.

src/sdk/http/http-client.ts
```diff
@@ -156,7 +158,34 @@ export abstract class HttpClient<TRequest, TResponse> implements IHttpClient {
 			}
 		});
 	}
-
+	
+	async download(path, handler): Promise<any> {
```
smbecker: This should be abstract and implemented in the `RequestHttpClient` because this still relies on implementation details that should not be a concern of the base HttpClient

src/sdk/http/http-client.ts
```diff
@@ -7,6 +7,8 @@ export interface IHttpClient {
 
 	get(path: string, params?: any): Promise<any>;
 
+	download(path: string, handler: any): Promise<any>;
```
smbecker: This should be `download(path: string, handler: (fileName: string, output: Readable)): Promise<any>;`

src/util/file-downloader.ts
```diff
@@ -0,0 +1,63 @@
...
+			} catch (err) {
+				if (err.code !== 'EEXIST') {
+					throw err;
+				}
+			}
+			return pathSegment;
+		}, path.isAbsolute(targetDirectory) ? path.sep : '');
+	}
+
+	private parseFileName(response: any): string {
```
smbecker: Add this as a protected method in the base http client.

src/sdk/http/request-client.ts
```diff
@@ -14,8 +14,22 @@ export class RequestHttpClient extends HttpClient<any, any> {
...
 		return response.statusCode;
 	}
+	
+	async download(path: string, handler: any) {
+		return await new Promise(async (resolve, reject) => {
+			try {
+				let options: any = await this.getOptions(path, {method: 'GET'});
+				let response = await request.get(options);
+				response.on('response', function (response) {
+					return resolve(handler(response));
```
smbecker:
```javascript
if (response.statusCode >= 400) {
	reject(new Error('The requested path could not be retrieved from the server.'));
	return;
}
return response(handler(this.parseContentDispositionHeader(response.headers), response));
```

src/sdk/http/request-client.ts
```diff
@@ -14,8 +14,22 @@ export class RequestHttpClient extends HttpClient<any, any> {
...
-
+	
 	protected getStatusCode(response: any): number {
 		return response.statusCode;
 	}
+	
+	async download(path: string, handler: any) {
+		return await new Promise(async (resolve, reject) => {
+			try {
+				let options: any = await this.getOptions(path, {method: 'GET'});
```
smbecker: These can be `const` since they are never changed.

PR #15, opened 2017-12-19
SF-1695 - Added agent sample script

smbecker:


samples/node/createAgent.js
```diff
@@ -0,0 +1,101 @@
...
+
+const client = new SkySyncClient(credentials);
+
+let nfsConnection, o365Connection, template, profile;
+
+async function addConnections() {
+	nfsConnection = await client.connections.add({
+		name: 'Sample NFS Connection',
+		platform: { id: 'nfs' },
+		disabled: true
```
recordrn: are the connections disabled because they only are needed for the template?

smbecker: exactly

PR #20, opened 2018-01-09
Fixed issue when auth token invalid

smbecker:


src/sdk/http/http-client.ts
```diff
@@ -255,6 +259,7 @@ export abstract class HttpClient<TRequest, TResponse> implements IHttpClient {
 
 								this.accessToken = null;
 								const token = await this.getAccessToken();
+								options.url = url;
```
smbecker: We need to re-add it because the `fetch` client will [explicitly delete](https://github.com/portalarchitects/skysync-cli/blob/master/src/sdk/http/fetch-client.ts#L40) it since its presence will cause issues with making the request.

PR #24, opened 2018-02-01
Sf 1629: Extend job item CLI commands to include csv export

rklind:


src/commands/jobs/items.ts
```diff
@@ -87,9 +96,16 @@ export = {
 					'status',
 					'processing',
 					'transferred_on'
-				]
-			});
-			output.writeTable(items, outputFormat);
+					]
```
smbecker: This bracket needs to be pulled in 1 tab.

src/sdk/resources/resource.ts
```diff
@@ -29,6 +29,7 @@ export function getPagedResponse<T>(response: any, items: T[]): PagedResult<T> {
 		hasMore: Boolean(meta && meta.has_more),
 		next: parseLink(meta && meta.links && meta.links.next),
 		previous: parseLink(meta && meta.links && meta.links.prev),
+		downloadLink: meta && meta.links && meta.links.download && meta.links.download.href,
```
smbecker: Should you just save all links rather than looking for this specific link? Especially since it is only valid for certain results.

src/commands/jobs/items.ts
rklind: will do!

src/sdk/resources/resource.ts
rklind: will do!

PR #25, opened 2018-02-02
SF-1326: Expose the management of profiles

rklind:


src/commands/profiles/add.ts
```diff
@@ -0,0 +1,78 @@
...
+			} ,
+			'templates': {
+				desc: 'A comma separated list of the job templates to associate to this new profile.  Both the ID and kind of the job template must be specified (ie. 123456:transfer).',
+				type: 'string'
+			}
+		});
+	},
+	handler: argv => {
+		runCommand(argv, async (client, output) => {
+			let newProfileRequestBody = {
```
smbecker: `const`

src/commands/profiles/download.ts
```diff
@@ -0,0 +1,27 @@
...
+				type: 'string',
+				default: (process.cwd())
+			}
+		})
+	},
+	handler: argv => {
+		runCommand(argv, async (client, output) => {
+			try {
+				const downloader = new FileDownloader(client.httpClient, client.profiles);
+				let path = await downloader.download(argv.id, argv.targetDirectory);
```
smbecker: `const`

src/commands/profiles/add.ts
```diff
@@ -0,0 +1,69 @@
...
+function isJobTemplateKindValid(jobTemplateKind) {
+	if (jobTemplateKinds.indexOf(jobTemplateKind) > -1) {
+		return true
+	}
+	return false;
+}
+
+function getJobTemplatesJson(argv, output) {
+	return argv.templates && argv.templates.split(',').map(input => {
+		const jobTemplateInputParsed = input.trim().split(':');
```
smbecker: Default to 'transfer' type, which will allow you to use destructuring...

`const [jobTemplateId, jobTemplateKind = 'transfer'] = input.trim().split(':');`

rklind: will do!

src/commands/profiles/add.ts
```diff
@@ -0,0 +1,61 @@
...
+]
+
+function isJobTemplateKindValid(jobTemplateKind) {
+	if (jobTemplateKinds.indexOf(jobTemplateKind) > -1) {
+		return true
+	}
+	return false;
+}
+
+function getJobTemplatesJson(argv, output) {
```
smbecker: Sorry, last thing... `output` is no longer used here.

rklind: will do, thanks!

PR #26, opened 2018-02-06
Sf 1327: Expose template endpoint

rklind:


src/commands/templates/util.ts
```diff
@@ -0,0 +1,52 @@
...
+			property: 'kind'
+		},
+		{
+			header: 'Enabled',
+			property: 'disabled',
+			transform: val => !val
+		}
+	],
+	json: [
+		'schedule'
```
smbecker: `transfer`

rklind: will do!

src/commands/templates/add.ts
```diff
@@ -0,0 +1,67 @@
...
+		runCommand(argv, async (client, output) => {
+			let template: any = {
+				kind: argv.kind,
+				name: argv.name,
+				schedule: {
+					mode: argv.manual ? 'manual' : 'auto'
+				}
+			};
+
+			let options = argv.optionsInput ? await readJsonInput() : undefined;
```
mikesae: We do the same option reading in the job, auditCategories, and Connections run command handlers. Consider Extract Method on that bit of code.

src/commands/templates/list.ts
```diff
@@ -0,0 +1,36 @@
...
+				desc: 'Only retrieve job templates of the specified kind',
+				type: 'string',
+				group: 'Search'
+			}, ...listArgumentsDefault
+		})
+	},
+	handler: argv => {
+		runCommand(argv, async (client, output) => {
+			const templates = await client.templates.list({
+				active: argv.active,
```
mikesae: use spread syntax to pass along argv properties:  `...argv`

src/commands/templates/add.ts
```diff
@@ -0,0 +1,67 @@
...
+				alias: 'file',
+				desc: 'The options JSON file',
+				type: 'string'
+			}
+		})
+	},
+	handler: argv => {
+		runCommand(argv, async (client, output) => {
+			let template: any = {
+				kind: argv.kind,
```
mikesae: use spread syntax to pass along argv properties:  `...argv`

smbecker: That will include a ton of additional parameters that are not related to the job, which we probably don't want. Especially since that would include authentication details.

src/sdk/models/templates.ts
```diff
@@ -0,0 +1,7 @@
+import { IEntityIdentifier } from './base';
+
+export interface Template extends IEntityIdentifier<string> {
```
smbecker: Templates should act on job interface. No need for the additional `Template` interface

smbecker:
Looks like this needs to be rebased as well

PR #28, opened 2018-02-10
SF-2018: Fix pools option so connections list even when its not specified.

mikesae:


smbecker:
I don't believe that this is the right fix. I'm guessing that this is related to the query string library change a few weeks back. It probably isn't excluding undefined values. That should be easily reproducible with an automated test.

mikesae:
Yes, the proper way to fix this is to not put undefined parameters in requests For example, `skysync-cli connections list` sends
```
GET http://localhost:9090/v1/connections?platform=&q=&active=&pools=&offset=0&limit=20&fields=name&fields=platform&fields=group&fields=pool
``` 
but it should send
```
GET http://localhost:9090/v1/connectionsoffset=0&limit=20&fields=name&fields=platform&fields=group&fields=pool
```
instead.

Fix [here](https://github.com/portalarchitects/skysync-cli/blob/master/src/sdk/http/http-client.ts#L107)  with something like 
```
import { isUndefined, pickBy} from 'lodash';
qs.stringify(_.pickBy(params, !_.isUndefined))
```

PR #29, opened 2018-02-15
SF-2018: Filter out undefined query parameters

rklind:


src/sdk/http/http-client.ts
```diff
@@ -99,12 +99,22 @@ export abstract class HttpClient<TRequest, TResponse> implements IHttpClient {
 		}
 	}
 
+	static getFilteredParameters(params?: any): any {
+		let filteredParams = {};
```
smbecker: `let filteredParams;`

src/sdk/http/http-client.ts
```diff
@@ -99,12 +99,22 @@ export abstract class HttpClient<TRequest, TResponse> implements IHttpClient {
...
+		for (const param in params) {
+			if (params[param] !== undefined) {
```
smbecker: Ensure `filteredParams` is defined here. `if (!filteredParams) { filteredParams = {}; }`

src/sdk/http/http-client.ts
```diff
@@ -99,12 +99,22 @@ export abstract class HttpClient<TRequest, TResponse> implements IHttpClient {
...
+			}
+		}
+		return filteredParams;
+	}
+
 	static getUrl(requestPath: string, baseUrl?: string, params?: any): string {
 		if (baseUrl && baseUrl.length > 0 && requestPath.indexOf('://') === -1) {
 			requestPath = baseUrl + requestPath;
 		}
 		if (params) {
```
smbecker: Change this to...

```
if (params) {
	params = HttpClient.getFilteredParameters(params);
}
if (params) {
	//qs.stringify
}
```

src/sdk/http/http-client.ts
```diff
@@ -99,12 +99,22 @@ export abstract class HttpClient<TRequest, TResponse> implements IHttpClient {
 		}
 	}
 
+	static getFilteredParameters(params?: any): any {
```
smbecker: I don't think this needs to be a `static` method attached to `HttpClient`. I would move it up so it isn't a part of the public api.

rklind: will do

src/sdk/http/http-client.ts
rklind: will do

src/sdk/http/http-client.ts
rklind: will do

src/sdk/http/http-client.ts
rklind: will do

src/sdk/http/http-client.ts
```diff
@@ -69,6 +69,19 @@ function invalidTokenError() {
 	return new Error('Could not acquire a valid access token.');
 }
 
+function getFilteredParameters(params?: any): any {
```
jtirjan: Method name isn't very descriptive. I'd suggest something more like `removeUndefinedParameters()`

rklind: will do, thanks!

PR #32, opened 2018-03-02
SF-2045 - Added mime_type and file_type properties to transfer item

smbecker:


src/sdk/models/util.test.ts
```diff
@@ -3,8 +3,8 @@ import expect = require('expect.js');
 
 describe('formatBytes', () => {
 	it('can format positive bytes', () => {
-		expect(util.formatBytes(10)).to.eql('10');
-		expect(util.formatBytes(100)).to.eql('100');
+		expect(util.formatBytes(10)).to.eql('10 B');
```
gotvino: Thanks for fixing this!

PR #34, opened 2018-03-15
Reduce unit formatting precision and optimize

smbecker:


src/sdk/models/util.ts
```diff
@@ -1,34 +1,42 @@
-const unitsAndFormatDigit = [
-	['B', 0],
-	['KB', 1],
-	['MB', 2],
-	['GB', 2],
-	['TB', 3]
-];
-const maxByteDigit = unitsAndFormatDigit.length - 1;
+const units = [ 'B', 'KB', 'MB', 'GB', 'TB', 'PB', 'EB', 'YB', 'ZB' ];
```
jtirjan: YB is greater than ZB

smbecker: Ha, so it is...good catch!

gotvino: ZBs? YBs? The future looks bright. 😂

PR #36, opened 2018-03-19
SF-1960 - Expose items through client

smbecker:


src/util/formatter.ts
```diff
@@ -34,6 +35,30 @@ export class OutputFormatter {
 		this.write(obj, options, true);
 	}
 
+	async writeAllPages<T>(func: (params: any) => Promise<PagedResult<T>>, params?: any, options?: OutputOptions): Promise<void> {
```
mikesae: `writeAllPages` doesn't seem to be used anywhere (I couldn't find any references to it).  If so, please remove.

src/sdk/http/http-client.test.ts
```diff
@@ -42,6 +42,16 @@ describe('HttpClient', () => {
...
 			expect(url).to.eql('http://localhost:8000/v1/connections?bar=barParameter&baz=bazParameter');
-		})
+		});
+		
+		it('should handle request path that has existing query string', () => {
+			const params = {
+				foo: undefined,
+				bar: 'barParameter',
+				baz: 'bazParameter'
+			}
```
mikesae: We should terminate the const with a semicolon here and in a few more places in this module.

smbecker: It looks like the lint settings need some love. I pushed `fix-lint` to update the lint settings and fix any issues. Once that is merged, I'll rebase this and fix any outstanding issues.

PR #38, opened 2018-04-16
SF-2211: modify job creation to accept json for the entire job creati

rklind:
…on block to support scheduling and personal_drive jobs

### DIFF
--- a/src/commands/jobs/add.ts
+++ b/src/commands/jobs/add.ts
@@ -55,7 +55,7 @@ export = {
 				...argv.name && { name: argv.name },
 				...argv.manual && { schedule: { mode: 'manual' }}
 			};
-			
+
 			let jobParametersInput = argv.jobInput ? await readJsonInput() : undefined;
 			if (!jobParametersInput && argv.jobFile) {
 				jobParametersInput = JSON.parse(fs.readFileSync(argv.jobFile, 'utf-8'));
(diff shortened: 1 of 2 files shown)


src/commands/jobs/add.ts
```diff
@@ -52,21 +50,25 @@ export = {
...
+			let jobParametersInput = argv.jobInput ? await readJsonInput() : undefined;
+			if (!jobParametersInput && argv.jobFile) {
+				jobParametersInput = JSON.parse(fs.readFileSync(argv.jobFile, 'utf-8'));
 			}
-			job[job.kind] = options;
 
-			if (!options && !argv.optionsInput && job.kind === 'transfer' && argv.wizard) {
+			let jobDefaults: any = {
+				kind: 'transfer',
+				schedule: { mode: 'auto' }
```
recordrn: I would suggest defaulting the mode to manual

rklind: the default mode was already auto.  i didn't change that as part of this code change.  is there a reason we should change from our standard?

recordrn: from what I recall, the jobs created via CLI where all manual (so maybe that original default was not working). Most use cases for creating jobs via scripting will be to create jobs ahead of time, but not executing them.  If this was set to 'auto' as a default, I would most likely override it and set to manual for the standard use scripts.

recordrn:
are we going to publish a new skysync-cli npm package?

smbecker:
We generally only do that with the next software release.

PR #39, opened 2018-04-30
SF-2231: skysync-cli cluster download

jtirjan:


src/commands/cluster/download.ts
```diff
@@ -0,0 +1,27 @@
...
+	},
+	handler: argv => {
+		runCommand(argv, async (client, output) => {
+			try {
+				const downloader = new FileDownloader(client.httpClient, client.cluster);
+				let path = await downloader.download(undefined, argv.targetDirectory);
+				output.writeSuccess(`File saved to ${path}`);
+			} catch (err) {
+				output.writeFailure(err);
+			}
```
rklind: this section of code looks about the same as two other similar files.  could this be abstracted and reused?

src/commands/cluster.ts
```diff
@@ -0,0 +1,5 @@
+export = {
+	command: 'cluster',
```
rklind: should this command be 'clusters' instead of 'cluster' to be consistent?

PR #50, opened 2018-07-12
SF-2548 - Add convention audit log resource

smbecker:


src/sdk/formatting/formatDate.ts
```diff
@@ -106,17 +108,21 @@ const defaultOptions = (options: DateFormatOptions): DateFormatOptions => {
 		options.now = new Date();
 	}
 
-	if (typeof(options.absolute) !== 'boolean') {
-		options.absolute = false;
+	if (typeof(options.allowRelative) !== 'boolean') {
```
rklind: do you have a test that tests that this is set to true when allowRelative is not of type boolean?

src/sdk/formatting/formatDateRange.ts
```diff
@@ -4,10 +4,10 @@ import { formatDate, DateFormatOptions } from './formatDate';
 
 export const formatDateRange = (start: ConvertibleDate, end: ConvertibleDate, options?: RangeFormatOptions & DateFormatOptions): string => {
 	options = options || {};
-	options.absolute = true;
+	options.allowRelative = false;
 
-	if (typeof(options.time) !== 'boolean') {
-		options.time = false;
+	if (typeof(options.displayTime) !== 'boolean') {
+		options.displayTime = false;
```
rklind: is there a test that tests this piece?

smbecker: Sort of since we don't define it in these tests: https://github.com/portalarchitects/skysync-cli/blob/sf-2548/src/sdk/formatting/formatDateRange.test.ts#L12

rklind: ...i guess.... ;)

PR #51, opened 2018-07-16
sf-2333: add support for transfer folders endpoint

rklind:


src/commands/jobs/util.ts
```diff
@@ -263,9 +263,9 @@ export const itemOutputFormat = {
 				if (!val) {
 					return '';
 				}
-				const size = !Boolean(val.source_to_destination)
-					? (val.source && val.source.bytes)
-					: (val.destination && val.destination.bytes);
+				const size = Boolean(val.source_to_destination) && val.destination && val.destination.bytes
+					? (val.destination && val.destination.bytes)
+					: (val.source && val.source.bytes);
```
rklind: does this change make sense or should i create a different output format for folder size? my concern with this logic is that if an item has 0 bytes on the destination, but it's not 0 bytes on the source, you'd see the source size, which may give a false impression of the contents on the destination...if there was a problem with the transfer.

src/commands/jobs/util.ts
```diff
@@ -263,9 +263,9 @@ export const itemOutputFormat = {
 				if (!val) {
 					return '';
 				}
-				const size = !Boolean(val.source_to_destination)
-					? (val.source && val.source.bytes)
-					: (val.destination && val.destination.bytes);
+				const size = Boolean(val.source_to_destination) && val.destination && val.destination.bytes
```
smbecker: I think that this statement should be included within the `Boolean` expression.

src/sdk/resources/transferFolders.ts
```diff
@@ -0,0 +1,17 @@
...
+	}
+	
+	async list(params?: any): Promise<TransferItem[]> {
+		this.resourcePath = `transfers/${params.id}/folders`;
+		delete params.id;
+		return super.list(params);
+	}
+}
+
+// const result = await this.httpClient.get(`transfers/${id}/folders`, this.mergeDefaultParams(params));
```
smbecker: ??

rklind: oops, thanks!

src/commands/jobs/util.ts
rklind: will do!

PR #56, opened 2018-07-25
Move platform comparison into CLI

smbecker:


package.json
```diff
@@ -54,6 +54,7 @@
     "cliff": "^0.1.10",
     "dot-object": "^1.7.0",
     "liftoff": "^2.5.0",
+    "lodash": "^4.17.10",
```
gotvino: Shouldn't package-lock.json be committed also? I still don't get why we even have them in github in the first place nor why they sometimes show up changed after an npm install.

PR #57, opened 2018-08-07
sf-2523 : cli platform comparison updates

recordrn:
Currently implemented ifFeaturePresentNotCompatible function for invalid_characters to make code freeze safely as a minimum requirement. Splitting off the array compare and modal popup. Note, restricted_type is currently unavailable on the platforms endpoint. Rule will be uncommented after defaults are set on the platform connectors.

### DIFF
--- a/src/sdk/models/platformCompare/rules.ts
+++ b/src/sdk/models/platformCompare/rules.ts
@@ -8,7 +8,8 @@ import {
 	ifSizeGreaterThan,
 	checkFeatures,
 	checkPath,
-	availableIfSupported
+	availableIfSupported,
+	ifStringArrayExists
 } from './util';
 
 const comparisonRules: PlatformComparisonRuleGroup[] = [
@@ -49,7 +50,7 @@ const comparisonRules: PlatformComparisonRuleGroup[] = [
 			{
 				id: 'invalid_characters',
 				name: 'Invalid characters',
-				compare: checkPath((left, right) => ifFeaturePresent(left, right, 'invalid_characters'))
+				compare: checkPath((left, right) => ifStringArrayExists(left, right, 'invalid_characters'))
 			},
 			{
 				id: 'max_file_size',
--- a/src/sdk/models/platformCompare/rules.ts
+++ b/src/sdk/models/platformCompare/rules.ts
@@ -80,19 +80,18 @@ const comparisonRules: PlatformComparisonRuleGroup[] = [
 		available: (left, right) => availableIfSupported(left, right, 'features.permissions'),
 		rules: [
 			{
-				id: 'author_preservation',
-				name: 'Author-Owner preservation',
-				compare: checkFeatures((left, right) => ifFeaturePresent(left, right, 'native_copy.ownership'))
-			},
-
-			{
 				id: 'map_accounts',
-				name: 'Map accounts',
+				name: 'Account map',
 				compare: checkFeatures((left, right) => ifFeaturePresent(left, right, 'query_account'))
 			},
 			{
+				id: 'author_preservation',
+				name: 'Author/Owner preservation',
+				compare: checkFeatures((left, right) => ifFeaturePresent(left, right, 'native_copy.ownership'))
+			},
+			{
 				id: 'map_groups',
-				name: 'Map groups',
+				name: 'Group map',
 				compare: checkFeatures((left, right) => ifFeaturePresent(left, right, 'query_group'))
 			},
 			{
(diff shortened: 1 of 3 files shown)


src/sdk/models/platformCompare/rules.ts
```diff
@@ -8,50 +8,118 @@ import {
...
+				name: 'Account map',
+				compare: checkFeatures((left, right) => ifFeaturePresent(left, right, 'query_account'))
 			},
 			{
-				id: 'max_file_size',
-				name: 'Maximum file size',
-				compare: checkFeatures((left, right) => ifSizeGreaterThan(left, right, 'max_size'))
+				id: 'author_preservation',
+				name: 'Author/Owner preservation',
+				compare: checkFeatures((left, right) => ifFeaturePresent(left, right, 'native_copy.ownership'))
```
smbecker: Not the right feature. Should be looking for `audit_trail`.

src/sdk/models/platformCompare/rules.ts
```diff
@@ -8,50 +8,118 @@ import {
...
 			}
 		]
 	},
 	{
-		id: 'lock_propagation',
-		name: 'Lock propagation',
-		available: (left, right) => availableIfSupported(left, right, 'features.checkout'),
+		id: 'metadata',
+		name: 'Metadata',
+		available: (left, right) => availableIfSupported(left, right, 'features.metadata'),
```
smbecker: Tags does not require metadata. I would make this an `OR`

src/sdk/models/platformCompare/util.test.ts
```diff
@@ -0,0 +1,19 @@
+import expect = require('expect.js');
+import {PlatformComparisonRuleStatus} from './types';
+import {
+	ifStringArrayExists
+} from './util';
+
+describe('ifStringArrayExists', () => {
+	it(' is Compatiable', () => {
```
smbecker: Preceding space not necessary.

src/sdk/models/platformCompare/util.ts
```diff
@@ -94,3 +95,30 @@ export const checkPath = (handler: (left: any, right: any) => PlatformComparison
 		return notApplicable();
 	};
 };
+
+export const ifStringArrayExists = (left: any, right: any, key: string): PlatformComparisonRuleResult => {
+	const leftStringArray = left === undefined || left instanceof Array ? left : Array(get(left, key));
+	const rightStringArray = right === undefined || right instanceof Array ? right : Array(get(right, key));
+	if (leftStringArray || rightStringArray) {
+		return {
+			left: leftStringArray || leftStringArray.length !== 0 ? leftStringArray.join(', ') : false,
```
smbecker: Should the CLI be responsible for concatenating the arrays? I would think that would be a UI concern.

src/sdk/models/platformCompare/rules.ts
```diff
@@ -8,50 +8,118 @@ import {
...
+	},
 	{
 		id: 'validation',
 		name: 'Path validation',
+		available: (left, right) => availableIfSupported(left, right, 'path.validation'),
 		rules: [
+			{
+				id: 'invalid_characters',
+				name: 'Invalid characters',
+				compare: checkPath((left, right) => ifFeaturePresentNotCompatible(left, right, 'invalid_characters'))
```
smbecker: Shouldn't this use `ifStringArrayExists`? I assume that is what the function is for and if the two are present but compatible (i.e. invalid characters from source are all considered invalid on destination) then we should show it as such.

recordrn: Yes. The intention is to use isStringArrayExists; however, I was running into an issue loading/implementing the `lodash.difference` function. Thought was to split out that functionality, and implement basic functionality as a stop gap for the code freeze.

src/sdk/models/platformCompare/util.ts
recordrn: fixed. removed `join`

src/sdk/models/platformCompare/util.test.ts
recordrn: fixed. removed leading spaces in `it` description

src/sdk/models/platformCompare/rules.ts
recordrn: fixed. changed to `audit_trail`

src/sdk/models/platformCompare/util.ts
```diff
@@ -101,8 +101,8 @@ export const ifStringArrayExists = (left: any, right: any, key: string): Platfor
 	const rightStringArray = right === undefined || right instanceof Array ? right : Array(get(right, key));
 	if (leftStringArray || rightStringArray) {
 		return {
-			left: leftStringArray || leftStringArray.length !== 0 ? leftStringArray.join(', ') : false,
-			right: rightStringArray || rightStringArray.length !== 0 ? rightStringArray.join(', ') : false,
+			left: leftStringArray || leftStringArray.length !== 0 ? leftStringArray : false,
```
smbecker: This should be an `&&`

recordrn: @smbecker - fixed.

PR #67, opened 2018-09-18
SF-2831 - Added request infrastructure for job user mapping data

krispeterson:


src/sdk/resources/transferSecurityMappings.ts
```diff
@@ -0,0 +1,15 @@
+import { IHttpClient } from '../http';
+import { getPagedResponse, getTypedResponse, PagedResource, PagedResult } from './resource';
+import { TransferSecurityMapping} from '../models';
+
+export class TransferSecurityMappingsResource extends PagedResource<TransferSecurityMapping> {
```
smbecker: I would just derive from `BaseResource` since there is nothing editable and no other valid method besides `page` and that is re-implemented below.

src/commands/jobs/util.ts
```diff
@@ -286,3 +286,24 @@ export const itemOutputFormat = {
 		'transferred_on'
 	]
 };
+
+export const securityMappingOutputFormat = {
+	table: [
+		{
+			header: 'Source',
+			property: 'source.email'
```
smbecker: `email` is not guaranteed to be the property that is present. I would set `property: 'source'` and use `transfer: val => val && (val.email || val.username || val.id)`

src/commands/jobs/util.ts
```diff
@@ -286,3 +286,24 @@ export const itemOutputFormat = {
...
+		},
+		{
+			header: 'Mapper',
+			property: 'mapper'
+		},
+		{
+			header: 'Message',
+			property: 'message'
+		}
+	]
```
smbecker: I would add `json: ['id', 'source', 'destination']`

src/commands/jobs/util.ts
```diff
@@ -286,3 +286,27 @@ export const itemOutputFormat = {
...
+			transform: val => val && (val.email || val.username || val.id)
+		},
+		{
+			header: 'Destination',
+			property: 'destination',
+			transform: val => val && (val.email || val.username || val.id)
+		},
+		{
+			header: 'Resolution',
+			property: 'mapper'
```
krispeterson: @avolinsky mentioned that he will be changing the `mapper` property name. I'll update it here as soon as his change is made.

src/sdk/resources/transferSecurityMappings.ts
krispeterson: Looks like `IDataStoreIdentifier` accepts `Resource` instead of `BaseResource` (https://github.com/portalarchitects/skysync-vnext/blob/6be097feef342df6b63a8bfd7b66086b110901a1/Hosting/web/app/data/stores/store.ts#L65). It also looks like our client code relies on some of the functions and properties defined in `Resource`. I'm going to refactor again to use `Resource` unless you know of another way around those dependencies.

krispeterson:
@smbecker looks like I need to rebase, but would you please take one last look now that the related vnext PR is approved.

krispeterson:
@smbecker, ping. Would you take one last look, please?

src/sdk/resources/transferSecurityMappings.ts
```diff
@@ -0,0 +1,15 @@
...
+import { getPagedResponse, getTypedResponse, Resource, PagedResult } from './resource';
+import { TransferSecurityMapping } from '../models';
+
+export class TransferSecurityMappingsResource extends Resource<TransferSecurityMapping> {
+	constructor(httpClient: IHttpClient) {
+		super(httpClient, 'item', 'item', null, null, 'transfers');
+	}
+
+	async page(params?: any): Promise<PagedResult<TransferSecurityMapping>> {
+		const result = await this.httpClient.get(`${this.resourcePath}/${params && params.id}/security_map`, this.mergeDefaultParams(params));
```
smbecker: This feels wrong but I suppose we can worry about fixing it once the services have support for exporting security maps for more than one job.

PR #71, opened 2018-09-24
SF-2737 - Fix audit_trail detection to include impersonation.per_request

smbecker:


smbecker:
In this particular case, it only needs one or the other.

PR #84, opened 2018-11-26
SF-1762 - Added job clone functionality

krispeterson:


src/commands/jobs/delete.ts
```diff
@@ -1,7 +1,7 @@
 import { runCommand } from '../../util/command';
 
 export = {
-	command: 'delete <id>',
+	command: 'delete [id]',
```
krispeterson: This was changed only for consistency, as was `show.ts` below. What is our standard? `[]` or `<>`?

smbecker: It [looks like](https://github.com/yargs/yargs/blob/master/docs/examples.md#yargs-is-here-to-help-you) `yargs` uses `<>` for command arguments and `[]` for option values.

krispeterson: Perfect. Those two changes should be valid then.

smbecker: Actually, based on that pattern, it should be `delete <id>` rather than `delete [id]`.

krispeterson: Updated this and others to use `<>` for command arguments. However, I did this only for those arguments which were required. For those arguments which appear optional, I left them qualified with `[]`. That may be splitting hairs, but that makes sense to me.

PR #85, opened 2018-11-28
SF-3136: Expose account and group exceptions and exclusions

rklind:


src/sdk/models/securityMaps.ts
```diff
@@ -0,0 +1,37 @@
...
+	};
+}
+
+export interface SecurityMapException {
+	source?;
+	destination?;
+}
+
+export interface SecurityMapExclusion {
+	source?: boolean;
```
smbecker: I don't believe that this is correct. I believe the correct property is `location?: 'source' | 'destination'`

PR #87, opened 2018-12-21
 Allow to patch multiple jobs

avolinsky:


src/sdk/resources/jobs.ts
```diff
@@ -27,6 +27,10 @@ export class JobsResource extends PagedResource<Job> {
 		return this.changeStatus(id, 'cancel', params);
 	}
 
+	patchMultiple(params: any, changes: Job): Promise<any> {
+		return this.httpClient.patch(`${this.resourcePath}`, changes, params);
```
smbecker: String interpolation is not necessary, just pass in `return this.httpClient.patch(this.resourcePath, changes, params);`

PR #88, opened 2018-12-21
SF-2927: EULA Acceptance

jtirjan:


src/sdk/resources/configuration.ts
```diff
@@ -0,0 +1,16 @@
+import { IHttpClient } from '../http';
+import { BaseResource } from './resource';
+
+export class ConfigurationResource extends BaseResource {
```
smbecker: Direct access to modify the config endpoints will not be available once an application is in `main` stage. We will likely want a dedicated endpoint for accepting the EULA

PR #91, opened 2019-01-22
SF-3529: Update copyright to SkySync

rklind:


LICENSE
```diff
@@ -1,6 +1,6 @@
 MIT License
 
-Copyright (c) 2017 Portal Architects, Inc.
+Copyright (c) 2017 SkySync, Inc.
```
jtirjan: Want to up the year?

rklind: should i?  i thought the copyright year is supposed to be when it was first made available...

PR #94, opened 2019-03-05
SF-3242: Add job convention to job model

rklind:


src/sdk/models/jobs.ts
```diff
@@ -58,6 +58,10 @@ export interface JobSchedule {
 	max_execution?: TimeInterval;
 }
 
+export interface JobConvention {
+	match?: string;
```
jtirjan: Should `match` be an enum?

src/sdk/models/jobs.ts
```diff
@@ -58,6 +58,16 @@ export interface JobSchedule {
...
 }
 
+export enum JobConventionMatch {
+	Account = 'account',
+	Container = 'container',
+	Ldap = 'ldap'
+}
+
+export interface JobConvention {
+	match?: JobConventionMatch;
```
smbecker: `match` is not a base property of convention options. If anything you would want `schedule` here.

rklind: i may not be understanding your comment.  can we touch base on this?  i was referencing the  documentation [here](https://skysync.atlassian.net/wiki/spaces/SSP/pages/13959311/User-Drive+Mapping+Feature).  also, my testing seemed to be working properly.

rklind: schedule for convention jobs is inside the transfer block.  inside the convention block is match, maby_by, remove_invalid, unmapped_policy, etc.

smbecker: but those are specific to the type of convention job. For example, unmapped_policy does not apply to folder mapping.

rklind: thanks for the chat @smbecker.  i get it now.  :)

PR #95, opened 2019-03-15
SF-3595: Add support for personal drive job inclusions and exclusions

rklind:


src/sdk/resources/personalDrive.ts
```diff
@@ -0,0 +1,13 @@
+import { IHttpClient } from '../http';
+import { PagedResource } from './resource';
+import { PersonalDriveMatchesResource } from './personalDriveMatches';
+
+export class PersonalDriveResource extends PagedResource<Account> {
+	constructor(httpClient: IHttpClient) {
+		super(httpClient, 'personal_drive');
+	}
+
+	matches(jobId: string): PersonalDriveMatchesResource {
```
smbecker: Might as well add exceptions to this as well

src/sdk/resources/personalDriveExceptions.ts
```diff
@@ -0,0 +1,9 @@
+import { IHttpClient } from '../http';
+import { PagedResource } from './resource';
+import { AccountMapException } from '../models/accountMaps';
+
+export class PersonalDriveExceptionsResource extends PagedResource<AccountMapException> {
+	constructor(httpClient: IHttpClient, personalDriveJobId: string) {
+		super(httpClient, 'item', 'item', 'item', 'item', `transfers/${personalDriveJobId}/personal_drives/exceptions`);
```
krispeterson: @rklind and @avolinsky, I'm sure you're already talking but just in case - looks like this and @avolinsky's [PR are duplicates](https://github.com/portalarchitects/skysync-cli/blob/57cca3a638efd7b0d5145e52e09a53e3b6d40dd5/src/sdk/resources/personalDriveExceptions.ts#L7).

avolinsky: yup, closed mine

src/sdk/resources/personalDrive.ts
```diff
@@ -0,0 +1,18 @@
...
+		return new PersonalDriveMatchesResource(this.httpClient, jobId);
+	}
+
+	exceptions(personalDriveJobId: string): PersonalDriveExceptionsResource {
```
smbecker: Why different parameter names for the same thing between this method and the one above?

rklind: copy and paste.  fix it now, though.

PR #99, opened 2019-04-11
SF-3855 | Added body param to DELETE request

avolinsky:


src/sdk/http/http-client.ts
```diff
@@ -34,7 +34,7 @@ export interface IHttpClient {
 
 	patch(path: string, body: any, params?: any): Promise<any>;
 
-	delete(path: string, params?: any): Promise<boolean>;
+	delete(path: string, body: any, params?: any): Promise<any>;
```
rklind: should body be optional?

rklind: should body be optional?

src/sdk/http/http-client.ts
```diff
@@ -376,10 +376,14 @@ export abstract class HttpClient<TRequest, TResponse> implements IHttpClient {
...
 	}
 
-	async delete(path: string, params?: any): Promise<boolean> {
-		const result = await this.executeApiRequest(path, params, {
-			method: 'DELETE'
-		});
-		return result !== null;
+	async delete(path: string, body: any, params?: any): Promise<any> {
+		if (body) {
+			return await this.executePost('DELETE', path, body, params);
```
rklind: will this line work even if body is null?

PR #100, opened 2019-05-07
SF-3464: Add Cancellation Token for Aborting Web Requests

jtirjan:


smbecker:
This feels like a bit of a leaky abstraction. It also shouldn't be limited to only GET requests. Plus this will not work with the `request` module as that uses a [different method](https://github.com/request/request/issues/772) for cancellation.

smbecker:
Should we return a custom `ResponsePromise<T>` object rather than `Promise<T>` and expose a `cancel` method? Or just handle it in the UI using a completely different means altogether by ignoring a stale response rather than explicitly aborting? Thoughts?

jtirjan:
There are some nice benefits to aborting stale requests. I'd prefer to go that route and expose a cancel method. I'll see what the LOE is on creating the custom response.

jtirjan:
I think I got them all. It was super fun.

PR #102, opened 2019-05-29
SF-2951: Add help_url to audit category model

jtirjan:


src/sdk/models/auditCategories.ts
```diff
@@ -1,7 +1,9 @@
 import { IEntityIdentifier } from './base';
 
 export interface AuditCategory extends IEntityIdentifier<string> {
+	id?: string;
```
smbecker: This shouldn't be necessary as the entity derives from `IEntityIdentifier` already.

jtirjan: rolled back

PR #104, opened 2019-06-14
SF-4087: Create job reset methods in jobs resource

rklind:


src/sdk/resources/jobs.ts
```diff
@@ -54,4 +54,53 @@ export class JobsResource extends PagedResource<Job> {
 		const result = await this.httpClient.patch(this.resourcePath, undefined, params, token);
 		return result;
 	}
+
+	resetValidate(id: string, start: boolean, params?: any, token?: CancellationToken): Promise<Job> {
```
smbecker: Would it be cleaner to have:

```tsx
enum JobResetType {
	validate = 'validate',
	validateFiltered = 'validate_filtered',
....
}

reset(id: string, reset: JobResetType = JobResetType.soft, params?: any, token?: CancellationToken): Promise<Job>
```

Also notice that I didn't include `start` parameter as you can always specify that in the `params` block.

PR #105, opened 2019-06-17
SF-3864: Add multi-item actions

jtirjan:


src/sdk/resources/transferItems.ts
```diff
@@ -16,11 +16,23 @@ export class TransferItemsResource extends PagedResource<TransferItem> {
 		return this.modifyItem(itemid, {retry: true}, token);
 	}
 
+	ignoreMultiple(itemids?: any, token?: CancellationToken): Promise<void> {
```
smbecker: Rename this to `params` since it wouldn't actually be a list of item IDs but rather our normal search parameters.

rklind: this is restricting us to providing a list of ids.  would it make more sense accept params here to allow for different search options?

PR #107, opened 2019-06-19
SF-4087: Expose execution resets via sdk

rklind:


src/sdk/models/jobs.ts
```diff
@@ -70,6 +82,7 @@ export interface Job extends IEntityIdentifier<string>, IAuditedEntity {
 	priority?: number;
 	status?: string;
 	execute_on?: number;
+	resets?: JobResetType[];
```
jtirjan: Should this exist at the job level, or could it be accessed via `previous_execution`?

rklind: this is for the next execution

PR #111, opened 2019-09-04
SF-3926: Expose the readonly property on the user object

rklind:


src/sdk/models/users.ts
```diff
@@ -10,8 +10,9 @@ export interface User extends IEntityIdentifier<string> {
 	name?: string;
 	email?: string;
 	disabled?: boolean;
+	readonly?: boolean;
```
jtirjan: Not sure if I'm missing something, but `SecurityUser` doesn't have a `readonly` property

smbecker: Ah nice catch. This should be modifying [Account](https://github.com/portalarchitects/skysync-cli/blob/master/src/sdk/models/connections.ts#L98) not `User`.

jtirjan: That inherits from `SecurityIdentifier`, which already has it - https://github.com/portalarchitects/skysync-cli/blob/master/src/sdk/models/connections.ts#L90

PR #120, opened 2019-12-03
SF-4102: Implement resetMultiple Jobs

jtirjan:


src/sdk/resources/jobs.ts
```diff
@@ -64,4 +64,14 @@ export class JobsResource extends PagedResource<Job> {
 		const job = await this.httpClient.patch(`${this.resourcePath}/${id}`, undefined, params, token);
 		return this.getSingle(job);
 	}
+
+	async resetMultiple(reset: JobResetType, params?: any, token?: CancellationToken): Promise<Job> {
```
smbecker: Couldn't this be simplified to

```typescript
resetMultiple(reset: JobResetType, params?: any, token?: CancellationToken): Promise<Job> {
	return this.httpClient.patch(`${this.resourcePath}`, undefined, this.mergeParams({reset}, params), token);
}
```

jtirjan: Will do. I just copied from the method above. I'll go ahead and update both.

PR #127, opened 2020-06-12
Vnext 110

rklind:


src/sdk/models/policies.ts
```diff
@@ -0,0 +1,18 @@
...
+	description?: string;
+	category?: JobCategory;
+	disabled?: boolean;
+}
+
+export interface PolicyJob extends Job {
+	policy?: PolicyJobOptions;
+}
+
+export interface PolicyJobOptions extends TransferPath {
```
smbecker, line 17: This should also include a list of `Policy` objects

src/sdk/models/policyStatistics.ts
```diff
@@ -0,0 +1,36 @@
...
+
+export interface PolicyStatistic {
+	name?: string;
+	description?: string;
+	bytes?: number;
+	files?: number;
+	folders?: number;
+	versions?: number;
+	storage?: number;
+	cost_savings?: number;
```
smbecker, line 11: Remove for now.

src/sdk/models/policyStatistics.ts
```diff
@@ -0,0 +1,36 @@
...
+export interface PolicyStatisticsList {
+	[key: string]: PolicyStatistic;
+}
+
+export type PolicyJobStatisticsTimeline = { 
+	timestamp: number; 
+	stats: {
+		by_risk?: PolicyStatisticsList;
+		by_group?: PolicyStatisticsList;
+		by_activity?: PolicyStatisticsList;
```
smbecker, line 23: `by_activity` will not be populated for phase 1. We will need to handle that correctly from the UI.

src/sdk/models/policyStatistics.ts
```diff
@@ -0,0 +1,36 @@
...
+	}
+}[];
+
+export interface PolicyJobStatistics extends IEntityIdentifier<string> {
+	timeline?: PolicyJobStatisticsTimeline;
+	by_risk?: PolicyStatisticsList;
+	by_group?: PolicyStatisticsList;
+	by_activity?: PolicyStatisticsList;
+	by_location_health?: PolicyStatisticsList;
```
smbecker, line 32: We won't have statistics by health.

rklind: How will the `locations` pie chart be populated in the `Policy overview`? https://www.figma.com/proto/54xvnM968kcVTAKfUTjc8X/Policy-Engine?node-id=174%3A751&scaling=min-zoom

smbecker: Ugh, I guess we will need to calculate then.

src/sdk/models/policies.ts
```diff
@@ -0,0 +1,19 @@
...
+	category?: PolicyCategory;
+	disabled?: boolean;
+}
+
+export interface PolicyJob extends Job {
+	policy?: PolicyJobOptions;
+}
+
+export interface PolicyJobOptions extends TransferPath {
+	policy?: Policy[];
```
smbecker, line 18:
```suggestion
	policies?: Policy[];
```

src/sdk/models/policyCategories.ts
```diff
@@ -0,0 +1,5 @@
+import { IEntityIdentifier } from './base';
+
+export interface PolicyCategory extends IEntityIdentifier<number> {
```
smbecker, line 3:
```suggestion
export interface PolicyCategory extends IEntityIdentifier<string> {
```

PR #131, opened 2020-06-25
VNEXT-114: Add policy audit endpoint

rklind:


src/sdk/models/policies.ts
```diff
@@ -14,5 +15,32 @@ export interface PolicyJob extends Job {
...
+}
+
+export interface PolicyItem extends IEntityIdentifier<number> {
+	parent_id?: number;
+	source?: TransferPlatformItem;
+	audit_category?: AuditCategory;
+	retried?: number;
+	status?: string;
+	processing?: string[];
+	source_to_destination?: boolean;
```
smbecker, line 29: No `source_to_destination`

src/sdk/models/policies.ts
```diff
@@ -14,5 +15,32 @@ export interface PolicyJob extends Job {
...
+	source?: TransferPath;
+}
+
+export interface PolicyItem extends IEntityIdentifier<number> {
+	parent_id?: number;
+	source?: TransferPlatformItem;
+	audit_category?: AuditCategory;
+	retried?: number;
+	status?: string;
+	processing?: string[];
```
smbecker, line 28: No `processing` for the time being.

src/sdk/models/policies.ts
```diff
@@ -14,5 +15,32 @@ export interface PolicyJob extends Job {
...
+	source_to_destination?: boolean;
+	transferred_on?: number;
```
smbecker, line 30: `processed_on`

src/sdk/models/policies.ts
```diff
@@ -1,7 +1,8 @@
 import { IEntityIdentifier, IAuditedEntity } from './base';
 import { JobCategory } from './jobCategories';
 import { Job } from './jobs';
-import { TransferPath } from './transfers';
+import { TransferPath, TransferPlatformItem } from './transfers';
+import { AuditCategory } from './auditCategories';
```
smbecker, line 5: We should refactor [this resource](https://github.com/portalarchitects/skysync-cli/blob/master/src/sdk/resources/auditCategories.ts#L5) since that is transfer specific and there will be policy-based audit categories coming [here](https://github.com/portalarchitects/skysync-vnext/pull/2458/files#diff-460d13e2b0c313d2ea00a49ec8e52b0d).

src/sdk/models/policies.ts
```diff
@@ -17,3 +18,27 @@ export interface PolicyJob extends Job {
 export interface PolicyJobOptions extends TransferPath {
 	policies?: Policy[];
 }
+
+export interface PolicyItem extends IEntityIdentifier<number> {
+	parent_id?: number;
+	source?: TransferPlatformItem;
```
smbecker, line 24: Sorry, I missed this but the tracking model for policy items will have these inlined rather than under the `source` prop.

PR #132, opened 2020-06-26
VNEXT-113: Add policy items resource

rklind:


src/sdk/models/policies.ts
```diff
@@ -26,10 +32,11 @@ export interface PolicyItem extends IEntityIdentifier<number> {
 	retried?: number;
 	status?: string;
 	processing?: string[];
-	source_to_destination?: boolean;
-	transferred_on?: number;
 	root?: boolean;
 	type?: 'container' | 'item';
+	risk_level?: PolicyItemRiskLevel;
+	tracking_group?: string;
+	activity?: string;
```
smbecker, line 39: `activity` will not be available in phase 1

src/sdk/resources/policyItemActivities.ts
```diff
@@ -0,0 +1,9 @@
+import { IHttpClient } from '../http';
+import { PagedResource } from './resource';
+import { PolicyItemActivity } from '../models/policies';
+
+export class PolicyItemActivitiesResource extends PagedResource<PolicyItemActivity> {
```
smbecker, line 5: There will be no activities endpoint in phase 1. That all still needs to be designed and fleshed out.

src/sdk/resources/policyItemTrackingGroups.ts
```diff
@@ -0,0 +1,9 @@
+import { IHttpClient } from '../http';
```
smbecker, line 1: Tracking groups would be retrieved from the policy `groups` property. (i.e. `/policies/{id}?fields=groups`)

src/sdk/resources/policies.ts
```diff
@@ -1,9 +1,14 @@
...
 import { PagedResource } from './resource';
 import { Policy } from '../models';
+import { PolicyItemTrackingGroupsResource } from './policyItemTrackingGroups';
 
 export class PoliciesResource extends PagedResource<Policy> {
 	constructor(httpClient: IHttpClient) {
 		super(httpClient, 'policy', 'policy', 'policies', 'policies', 'policies');
 	}
+
+	trackingGroups(policyId: string): PolicyItemTrackingGroupsResource {
```
smbecker, line 11: This can be removed.

rklind: I was using this to set the text on the `Tracking group` column on the items grid. Otherwise, I'd have the tracking group id on the item, right?

src/sdk/resources/policyItemActivities.ts
rklind: I was using this to set the text on the `Activities` column on the items grid. Otherwise, I'd have the activity id on the item, right? Should I remove this column from the items grid (for phase 1)?

smbecker: Yes

src/sdk/resources/policies.ts
smbecker: The item will have a `group.id` and `group.name` attached to the item so you can grab the text from that.

src/sdk/models/policies.ts
```diff
@@ -27,6 +33,8 @@ export interface PolicyItem extends TransferPlatformItem {
 	processed_on?: number;
 	root?: boolean;
 	type?: 'container' | 'item';
+	risk_level?: PolicyItemRiskLevel;
```
smbecker, line 36: `risk_level` will be within the tracking group

src/sdk/models/policies.ts
```diff
@@ -27,6 +33,8 @@ export interface PolicyItem extends TransferPlatformItem {
...
+	tracking_group?: PolicyItemTrackingGroup;
```
smbecker, line 37:
```suggestion
	group?: PolicyItemTrackingGroup;
```

src/sdk/models/policies.ts
```diff
@@ -41,3 +49,7 @@ export interface PolicyAuditEntry extends IEntityIdentifier<number> {
 	message?: string;
 	recorded_on?: number;
 }
+
+export interface PolicyItemTrackingGroup extends IEntityIdentifier<number> {
```
smbecker, line 53:
```suggestion
export interface PolicyItemTrackingGroup extends IEntityIdentifier<string> {
```

src/sdk/models/policies.ts
rklind: The tracking group has a risk level associated with it?

PR #134, opened 2020-07-16
SF-4523 add discriminator field to jobs list command

bentatum:


src/commands/jobs/util.ts
```diff
@@ -53,6 +53,10 @@ export const listOutputFormat = {
 		{
 			header: 'Status',
 			property: 'status'
+		},
+		{
+			header: 'Discriminator',
+			property: 'discriminator'
```
rklind, line 59: Since `discriminator` is so closely related to `name`, maybe it makes sense to order it that way in the table.

bentatum: Good point, I adjusted it.

PR #139, opened 2020-09-01
ACT-29 | Support upload of form data directly

gotvino:


gotvino:
I will add these @smbecker. Do we, however, want to also add policy command support? Because that would be a bit of an undertaking, outside the scope of this card. That could have been another approach to this demo, actually; would have been easier to "install".

smbecker:
> Do we, however, want to also add policy command support?

Not at this point. As you said, that would be outside of the scope of this card.

PR #143, opened 2020-09-22
Sf-6913 | Catch up to future design system upgrades

MicheleBuselli:
- A few patch upgrades to match versions that will be referenced by design system upgrades
- Also a couple minor upgrades that are new-ish

Note that 1 test (formatDate should format days-distant dates) fails for me locally but passes in Azure.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -48,9 +48,9 @@
       "dev": true
     },
     "@types/node": {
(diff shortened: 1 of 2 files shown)


jeffrimko (approved):
Quick visual review, didn't test on my local. Looks good.

PR #153, opened 2020-12-15
SF-7686 | Add rendition to TransferOptions and create RenditionType enum

bentatum:
related: https://github.com/portalarchitects/skysync-vnext/pull/2875

### DIFF
--- a/src/sdk/models/transfers.ts
+++ b/src/sdk/models/transfers.ts
@@ -64,6 +64,11 @@ export enum TransferType {
 	Taxonomy = 'taxonomy'
 }
 
+export enum RenditionType {
+	Rendition = "rendition",
+	Original = "original"
+}
+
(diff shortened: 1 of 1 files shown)


src/sdk/models/transfers.ts
```diff
@@ -64,6 +64,11 @@ export enum TransferType {
 	Taxonomy = 'taxonomy'
 }
 
+export enum RenditionType {
```
smbecker, line 67: Should we make [naming consistent](https://github.com/portalarchitects/skysync-vnext/blob/bd48a7a2dd008c0d8a0c4b00ae94f8ad49a3f5c2/TransferEngine/src/PortalArchitects.Transfers.Primitives/RenditionSelectionMethod.cs#L9)?

smbecker (approved):

PR #154, opened 2020-12-16
ACT-534: Return Policies After Import

jtirjan:


src/sdk/resources/policies.ts
```diff
@@ -8,8 +8,8 @@ export class PoliciesResource extends PagedResource<Policy> {
 		super(httpClient, 'policy', 'policy', 'policies', 'policies', 'policies');
 	}
 
-	async import(body: FormData, params?: any, token?: CancellationToken): Promise<void> {
-		await this.httpClient.upload(this.resourcePath, null, body, params, token);
+	async import(body: FormData, params?: any, token?: CancellationToken): Promise<Policy[]> {
+		return await this.httpClient.upload(this.resourcePath, null, body, params, token);
```
smbecker, line 12: Doesn't this need to be wrapped in a `getTypedResponse(await this.httpClient.upload(this.resourcePath, null, body, params, token));`?

smbecker (changes requested):
Might as well update [entity types](https://github.com/portalarchitects/skysync-cli/blob/7e5dd5d5f8f46ecfff34d5dc234cfcd9e064966a/src/sdk/resources/entityTypes.ts#L11) as well.

jtirjan:
> Might as well update [entity types](https://github.com/portalarchitects/skysync-cli/blob/7e5dd5d5f8f46ecfff34d5dc234cfcd9e064966a/src/sdk/resources/entityTypes.ts#L11) as well.

I can do that, but I'll need to update that endpoint as well

smbecker:
> > Might as well update [entity types](https://github.com/portalarchitects/skysync-cli/blob/7e5dd5d5f8f46ecfff34d5dc234cfcd9e064966a/src/sdk/resources/entityTypes.ts#L11) as well.
> 
> I can do that, but I'll need to update that endpoint as well

Should have been [updated already](https://github.com/portalarchitects/skysync-vnext/blob/act-719/PolicyEngine/src/PortalArchitects.Policies.Services/V1/EntityTypesController.cs#L178). If something was missed, please add a comment to the PR.

smbecker (approved):

PR #156, opened 2020-12-17
Upgrade and trim dependencies

smbecker:


gotvino (approved):

gotvino (approved):
Gave it a quick spin and it works fine, at least connections and jobs

PR #160, opened 2021-01-22
Act-625: UI | Policy creation | Labels view | Filters

mmitoff:
Models necessary for ACT-625


src/sdk/models/metadataFilters.ts
```diff
@@ -0,0 +1,86 @@
+export enum MetadataConjunctionOperator {
+	And = 'and',
+	Or = 'or'
+}
+
+export enum MetadataPropertyFilterOperator {
```
smbecker, line 6: This doesn't look right and doesn't match the [server-side operators](https://github.com/portalarchitects/skysync-vnext/blob/198997fbe99590622e09930c2ad7572b52f288d7/Connectors/Connectors/src/PortalArchitects.Connectors.Primitives/Metadata/MetadataPropertyFilterOperator.cs#L11).

src/sdk/models/policies.ts
```diff
@@ -73,7 +74,18 @@ export interface PolicyAuditEntry extends IEntityIdentifier<number> {
...
+}
+
 export interface PolicyTrackingGroup extends IEntityIdentifier<string>, IPrioritizedEntity {
 	name?: string;
+	description?: string;
+	priority?: number;
 	risk?: ContentRiskLevel;
+	rules?: any[];
+	assignment_rules?: PolicyTrackingGroupRule[];
+	entity_types?: any[];
```
smbecker, line 89:
```suggestion
	entity_types?: (EntityType & IPrioritizedEntity)[];
```

src/sdk/models/policies.ts
```diff
@@ -73,7 +74,18 @@ export interface PolicyAuditEntry extends IEntityIdentifier<number> {
...
+	justification?: string;
+	criteria?: MetadataCalculatedFilter;
+}
+
 export interface PolicyTrackingGroup extends IEntityIdentifier<string>, IPrioritizedEntity {
 	name?: string;
+	description?: string;
+	priority?: number;
 	risk?: ContentRiskLevel;
+	rules?: any[];
```
smbecker, line 87:
```suggestion
```

src/sdk/models/policies.ts
```diff
@@ -73,7 +74,18 @@ export interface PolicyAuditEntry extends IEntityIdentifier<number> {
...
 
+export interface PolicyTrackingGroupRule extends IEntityIdentifier<string> {
+	justification?: string;
+	criteria?: MetadataCalculatedFilter;
+}
+
 export interface PolicyTrackingGroup extends IEntityIdentifier<string>, IPrioritizedEntity {
 	name?: string;
+	description?: string;
+	priority?: number;
```
smbecker, line 85: Remove as this property comes through with `IPrioritizedEntity`.

smbecker (changes requested):

src/sdk/models/metadataFilters.ts
```diff
@@ -0,0 +1,86 @@
...
+}
+
+export interface MetadataPropertyFilter extends MetadataComponentFilter {
+	name?: string;
+	op?: MetadataPropertyFilterOperator;
+	value?: object;
+	select?: MetadataCalculatedExpression;
+}
+
+export interface MetadataScriptFilter extends MetadataComponentFilter {
```
smbecker, line 60: Remove as we don't actively support this at the moment

src/sdk/models/metadataFilters.ts
```diff
@@ -0,0 +1,86 @@
...
+	In = 'In',
+	not_in = 'Not In',
+	NotIn = 'Not In',
+	empty = 'Empty',
+	Empty = 'Empty',
+	not_empty = 'Not Empty',
+	NotEmpty = 'Not Empty'
+}
+
+export interface MetadataComponentFilter { }
```
smbecker, line 41:
```suggestion
export type MetadataComponentFilter = MetadataConjunctionFilter | MetadataPropertyFilter;
```

Then remove the `extends MetadataComponentFilter` on the various filters

src/sdk/models/metadataFilters.ts
```diff
@@ -0,0 +1,86 @@
...
+
+export interface MetadataCalculatedFilter {
+	text: string;
+	expression: MetadataComponentFilter;
```
smbecker, line 45:
```suggestion
	text?: string;
	expression?: MetadataComponentFilter;
```

src/sdk/models/metadataFilters.ts
```diff
@@ -0,0 +1,86 @@
...
+	op?: MetadataPropertyFilterOperator;
+	value?: object;
+	select?: MetadataCalculatedExpression;
+}
+
+export interface MetadataScriptFilter extends MetadataComponentFilter {
+	script: string;
+}
+
+export interface MetadataComponentExpression { }
```
smbecker, line 64:
```suggestion
export type MetadataComponentExpression = MetadataFunctionExpression | MetadataPropertyExpression MetadataStaticExpression;
```

src/sdk/models/metadataFilters.ts
```diff
@@ -0,0 +1,86 @@
...
+export interface MetadataFunctionExpression extends MetadataComponentExpression {
+	function: string;
+	args: MetadataComponentExpression[];
+}
+
+export interface MetadataPropertyExpression extends MetadataComponentExpression {
+	name: string;
+}
+
+export interface MetadataScriptExpression extends MetadataComponentExpression {
```
smbecker, line 80: Remove

src/sdk/models/metadataFilters.ts
```diff
@@ -0,0 +1,86 @@
...
+export interface MetadataComponentExpression { }
+
+export interface MetadataCalculatedExpression {
+	text: string;
+	expression: MetadataComponentExpression;
+}
+
+export interface MetadataFunctionExpression extends MetadataComponentExpression {
+	function: string;
+	args: MetadataComponentExpression[];
```
smbecker, line 73:
```suggestion
	args?: MetadataComponentExpression[];
```

src/sdk/models/metadataFilters.ts
```diff
@@ -0,0 +1,86 @@
...
+
+export interface MetadataScriptFilter extends MetadataComponentFilter {
+	script: string;
+}
+
+export interface MetadataComponentExpression { }
+
+export interface MetadataCalculatedExpression {
+	text: string;
+	expression: MetadataComponentExpression;
```
smbecker, line 68:
```suggestion
	text?: string;
	expression?: MetadataComponentExpression;
```

src/sdk/models/metadataFilters.ts
```diff
@@ -0,0 +1,86 @@
...
+export interface MetadataPropertyExpression extends MetadataComponentExpression {
+	name: string;
+}
+
+export interface MetadataScriptExpression extends MetadataComponentExpression {
+	script: string;
+}
+
+export interface MetadataStaticExpression extends MetadataComponentExpression {
+	value: string;
```
smbecker, line 85:
```suggestion
	value: any;
```

smbecker (changes requested):

src/sdk/models/metadataFilters.ts
mmitoff: I began with just the server side enum values for interpreting the "op" values on the metadata property filters.  I ended up adding the enum members as values as well to handle interpreting the "function" values on metadata function expressions. I could move the enum member values "Equals, NotEqual, etc" into a different enum?

smbecker: This enum should just be the server side enum values. Anything else should be a separate enum. Also, I don't recall us having function names like 'Equals' and 'NotEquals'. The possible function names comes from the registered function provides, which is defaulted [here](https://github.com/portalarchitects/skysync-vnext/blob/develop/Connectors/Connectors/src/PortalArchitects.Connectors.Transfers/Metadata/MetadataExpressionParser.cs#L28). You could make the default functions into an enum and then change `MetadataFunctionExpression.name` to `MetadataFunctionName | string`.

mmitoff: Refactored

src/sdk/models/metadataFilters.ts
```diff
@@ -0,0 +1,81 @@
+export enum MetadataConjunctionOperator {
+	And = 'and',
+	Or = 'or'
+}
+
+export enum MetadataPropertyFilterOperator {
+	eq = 'Equals',
```
smbecker, line 7: These should actually be reversed (i.e. `Equals = 'eq'`).

src/sdk/models/metadataFilters.ts
```diff
@@ -0,0 +1,81 @@
...
+	empty = 'Empty',
+	not_empty = 'Not Empty',
+}
+
+export enum MetadataFunctionName {
+	Now = 'Now',
+	Year = 'Year',
+	Month = 'Month',
+	Day = 'Day',
+	AddDays = 'Add Days',
```
smbecker, line 30: The part in quotes should match the function name, which shouldn't have any spaces in it.

smbecker (changes requested):

src/sdk/models/metadataFilters.ts
mmitoff: Should I update the enum member name as well then (change 'Less Than' or LessThan) or leave as is:

export enum MetadataPropertyFilterOperator {
	Equals = 'eq',
	'Not Equal'= 'ne',
	'Less Than' = 'lt',
	'Less Than Equal To' = 'le',
	'Greater Than' = 'gt',
	'Greater Than Equal To' = 'ge',
	'Starts With' = 'starts_with',
	'Does Not Start With' = 'not_starts_with',
	'Ends With' = 'ends_with',
	'Does Not End With' = 'not_ends_with',
	Contains = 'contains',
	'Does Not Contain' = 'not_contains',
	In = 'in',
	'Not In' = 'not_in',
	Empty = 'empty',
	'Not Empty' = 'not_empty',
}

mmitoff: Updated

src/sdk/models/metadataFilters.ts
mmitoff: Updated

smbecker (approved):

PR #165, opened 2021-02-18
ACT-626: PolicyTrackingGroupAction Model

jtirjan:


src/sdk/models/policies.ts
```diff
@@ -79,11 +79,23 @@ export interface PolicyTrackingGroupRule extends IEntityIdentifier<string> {
 	criteria?: MetadataCalculatedFilter;
 }
 
+export interface PolicyTrackingGroupActionConfiguration {
+	id?: string;
+	name?: string;
+	description?: string;
+}
```
smbecker, line 86:
```suggestion
export interface PolicyActionDescriptor extends IEntityIdentifier<string> {
	name?: string;
	description?: string;
}

export interface PolicyTrackingGroupActionConfiguration extends PolicyActionDescriptor {
	[name: string]: any;
}
```

src/sdk/models/policies.ts
```diff
@@ -79,11 +79,23 @@ export interface PolicyTrackingGroupRule extends IEntityIdentifier<string> {
...
+
+export interface PolicyTrackingGroupAction extends IEntityIdentifier<string> {
+	id?: string;
```
smbecker, line 89: Not necessary since it already derives from `IEntityIdentifier<string>`

```suggestion
```

smbecker (changes requested):

smbecker (approved):

PR #167, opened 2021-02-26
SF-8245 | remove map_by and unmapped_policy from PersonalDriveMapping…

bentatum:
…Convention interface + delete personalDriveExceptions.ts


JimPartinSkySync:
I think this should be removed (in src/sdk/resources/index.ts): https://github.com/portalarchitects/skysync-cli/blob/1b4538c9be08e7fc36646e58921298ae3f46999d/src/sdk/resources/index.ts#L26

JimPartinSkySync (changes requested):
I think this should be removed (in src/sdk/resources/index.ts):

https://github.com/portalarchitects/skysync-cli/blob/1b4538c9be08e7fc36646e58921298ae3f46999d/src/sdk/resources/index.ts#L26

JimPartinSkySync (approved):

avolinsky (approved):

PR #170, opened 2021-03-16
ACT:551 | Adding typings for PolicyAuditEntry

mmitoff:
Adding typings for PolicyAuditEntry

### DIFF
--- a/src/sdk/models/policies.ts
+++ b/src/sdk/models/policies.ts
@@ -63,9 +63,43 @@ export interface PolicyItem extends TransferPlatformItem {
 	group?: PolicyTrackingGroup;
 }
 
+export enum PolicyAuditEvent {
+	None = 'none',
+	Trace = 'trace',
+	Verbose = 'debug',
+	Information = 'info',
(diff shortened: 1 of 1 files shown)


src/sdk/models/policies.ts
```diff
@@ -63,9 +63,43 @@ export interface PolicyItem extends TransferPlatformItem {
 	group?: PolicyTrackingGroup;
 }
 
+export enum PolicyAuditEvent {
```
smbecker, line 66: Rather than having this type here, I think that we should stick to a mapping similar to what is in [core](https://github.com/portalarchitects/skysync-vnext/blob/develop/Hosting/web/app/pages/Jobs/View/ActivityLog/util.tsx#L17).

smbecker (changes requested):

src/sdk/models/policies.ts
mmitoff: Yeah that works, once you showed me that mapping this typing became obsolete. I'll close this PR

PR #180, opened 2021-06-02
ACT-1163: Add PolicyTrackingGroup Resource

jtirjan:


src/sdk/resources/policyTrackingGroups.ts
```diff
@@ -0,0 +1,15 @@
+import { IHttpClient } from '../http';
+import { PagedResource}  from './resource';
+import { PolicyTrackingGroup } from '../models';
+
+export class PolicyTrackingGroupsResource extends PagedResource<PolicyTrackingGroup> {
+	constructor(httpClient: IHttpClient, policyId?: string) {
```
smbecker, line 6: This should be a similar design to [here](https://github.com/portalarchitects/skysync-cli/blob/e022338957ee539ca837cbacac69148d65cedff4/src/sdk/resources/accountMaps.ts#L12) where `policyId` is required and this resource hangs off of the `PolicyResource`

smbecker (changes requested):

src/sdk/resources/policyTrackingGroups.ts
jtirjan: Ah! I knew there had to be another place where we were doing this. Thanks!

smbecker: Also `policies/tracking_groups` is not a valid route, even after https://github.com/portalarchitects/skysync-vnext/pull/3261 merges

jtirjan: It was recently. @mmitoff just made a change to use `/policies/{id}/tracking_groups` for all methods

smbecker (approved):

PR #185, opened 2021-06-11
ACT-1178 | adding entity type validators resource

bentatum:


gotvino (commented):
Don’t you need a change to [client](src/sdk/client.ts) to expose the validators rest entry point?

src/sdk/models/validators.ts
```diff
@@ -0,0 +1,6 @@
+import { IEntityIdentifier } from './base';
+
+export interface Validator extends IEntityIdentifier<string> {
```
smbecker, line 3: For consistency, I would change this to `EntityTypeValidator`

smbecker (changes requested):

smbecker (approved):

PR #187, opened 2021-06-14
CL99-63 | List users not filtering by active

DinuAlin:
- Changed active property to status

### DIFF
--- a/src/commands/users/list.ts
+++ b/src/commands/users/list.ts
@@ -8,7 +8,7 @@ export = {
 	builder: yargs => {
 		yargs.options({
 			'status': {
-				desc: 'Retrieves users acording to the given status. Status can only be "active" or "disabled"',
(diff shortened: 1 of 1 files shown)


src/commands/users/list.ts
```diff
@@ -7,9 +7,9 @@ export = {
 	desc: 'List all users',
 	builder: yargs => {
 		yargs.options({
-			'active': {
-				desc: 'Only retrieve active users',
-				type: 'boolean',
+			'status': {
+				desc: 'Retrieves users acording to the given status. Status can only be "active" or "disabled"',
```
MicheleBuselli, line 11: FIx spelling 'according'

MicheleBuselli (changes requested):

src/commands/users/list.ts
DinuAlin: Done

MicheleBuselli (approved):

PR #188, opened 2021-06-14
ACT-1167 - Add schema support to policies and entity types

smbecker:


mmitoff (approved):

src/sdk/resources/policies.ts
```diff
@@ -13,6 +13,11 @@ export class PoliciesResource extends PagedResource<Policy> {
 		return new PolicyTrackingGroupsResource(this.httpClient, policyId);
 	}
 
+	async schema(id: string, token?: CancellationToken): Promise<PropertySchema> {
+		const result = await this.httpClient.get(`${this.resourcePath}/${id}/schema`, null, token);
```
jtirjan, line 17: Passing `null` for `params` causes `http-client` to bomb

```suggestion
		const result = await this.httpClient.get(`${this.resourcePath}/${id}/schema`, {}, token);
```

image

jtirjan (changes requested):

src/sdk/resources/policies.ts
jtirjan: When attempting to destructure `params`
image

src/sdk/resources/entityTypes.ts
```diff
@@ -13,6 +13,11 @@ export class EntityTypesResource extends PagedResource<EntityType | BlockListEnt
 		return getTypedResponse<EntityType[]>(result);
 	}
 
+	async schema(id: string, token?: CancellationToken): Promise<PropertySchema> {
+		const result = await this.httpClient.get(`${this.resourcePath}/${id}/schema`, null, token);
```
jtirjan, line 17: This one as well
```suggestion
		const result = await this.httpClient.get(`${this.resourcePath}/${id}/schema`, {}, token);
```

src/sdk/resources/policies.ts
smbecker: Should be good now.

jtirjan (approved):

PR #191, opened 2021-06-24
ACT-1169 | UI | Policy Editor | Allow managing list of actions

jeffrimko:
Added new PolicyActionsResource and associated PolicyAction model so the app can retrieve the available actions from the `/v1/policies/actions` endpoint.

### DIFF
--- a/src/sdk/models/policies.ts
+++ b/src/sdk/models/policies.ts
@@ -78,12 +78,6 @@ export interface PolicyAuditEntry extends IEntityIdentifier<number> {
 	recorded_on?: number;
 }
 
-export interface PolicyAction extends IEntityIdentifier<string> {
-	name?: string;
-	description?: string;
-	destructive?: boolean;
-}
-
 export interface PolicyTrackingGroupRule extends IEntityIdentifier<string> {
 	justification?: string;
 	criteria?: MetadataCalculatedFilter;
@@ -92,6 +86,11 @@ export interface PolicyTrackingGroupRule extends IEntityIdentifier<string> {
 export interface PolicyActionDescriptor extends IEntityIdentifier<string> {
 	name?: string;
 	description?: string;
+	destructive?: boolean;
+	exclusive?: boolean;
+	requires_existing?: boolean;
+	removes_item?: boolean;
+	prompt?: any;
 }
 
 export interface PolicyTrackingGroupActionConfiguration extends PolicyActionDescriptor {
--- a/src/sdk/models/policies.ts
+++ b/src/sdk/models/policies.ts
@@ -78,6 +78,12 @@ export interface PolicyAuditEntry extends IEntityIdentifier<number> {
 	recorded_on?: number;
 }
 
+export interface PolicyAction extends IEntityIdentifier<string> {
+	name?: string;
+	description?: string;
+	destructive?: boolean;
+}
+
 export interface PolicyTrackingGroupRule extends IEntityIdentifier<string> {
 	justification?: string;
 	criteria?: MetadataCalculatedFilter;
@@ -92,8 +98,7 @@ export interface PolicyTrackingGroupActionConfiguration extends PolicyActionDesc
 	[name: string]: any;
 }
 
-export interface PolicyTrackingGroupAction extends IEntityIdentifier<string> {
-	priority?: number;
+export interface PolicyTrackingGroupAction extends IEntityIdentifier<string>, IPrioritizedEntity {
 	action?: PolicyTrackingGroupActionConfiguration;
 }
 
--- a/src/sdk/models/policies.ts
+++ b/src/sdk/models/policies.ts
@@ -83,9 +83,12 @@ export interface PolicyTrackingGroupRule extends IEntityIdentifier<string> {
 	criteria?: MetadataCalculatedFilter;
 }
 
-export interface PolicyActionDescriptor extends IEntityIdentifier<string> {
+interface PolicyActionBase extends IEntityIdentifier<string> {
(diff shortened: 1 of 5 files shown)


src/sdk/models/policies.ts
```diff
@@ -78,6 +78,12 @@ export interface PolicyAuditEntry extends IEntityIdentifier<number> {
 	recorded_on?: number;
 }
 
+export interface PolicyAction extends IEntityIdentifier<string> {
+	name?: string;
+	description?: string;
+	destructive?: boolean;
+}
```
jtirjan, line 85: This looks like a duplicate of `PolicyActionDescriptor`

src/sdk/resources/policyActions.ts
```diff
@@ -0,0 +1,9 @@
+import { IHttpClient } from '../http';
+import { PagedResource } from './resource';
+import { PolicyAction } from '../models';
+
+export class PolicyActionsResource extends PagedResource<PolicyAction> {
+	constructor(httpClient: IHttpClient) {
+		super(httpClient, 'policy_action', 'policy_action', 'policy_actions', 'policy_actions', 'policies/actions');
+	}
+}
```
jtirjan, line 9: Given that there is no direct link between Policies and Actions, the name `PolicyAction` and the route `policies/actions` are a little confusing.
Should we rename `PolicyActions` to something more accurate?
Let's PL tomorrow.

mmitoff: In vnext maybe the Actions controller endpoints move into the TrackingGroupsController since these are technically group actions? Definitely a good PL topic for tomorrow though.

src/sdk/models/policies.ts
mmitoff: Looks like PolicyActionDescriptor is missing some properties as well, might want to add exclusive, destructive, requires_existing, removes_item, and prompt to PolicyActionDescriptor after removing this.

jeffrimko: Yeah, they are similar currently and it might make sense to combine them. Thought it might make sense to separate them at first, i.e. the policy action details provided by the `/v1/policies/actions` endpoint and the policy action details once assigned to a tracking group. The case for keeping them separate would be that the unassigned action details might contain information that is not needed once assigned to a tracking group, e.g. the destructive flag or the exclusive flag.

src/sdk/models/policies.ts
```diff
@@ -86,14 +86,18 @@ export interface PolicyTrackingGroupRule extends IEntityIdentifier<string> {
 export interface PolicyActionDescriptor extends IEntityIdentifier<string> {
 	name?: string;
 	description?: string;
+	destructive?: boolean;
+	exclusive?: boolean;
+	requires_existing?: boolean;
+	removes_item?: boolean;
+	prompt?: any;
```
jeffrimko, line 93: Figured this would eventually change to a specific type but leaving as `any` for now since it's out of scope of the current card.

src/sdk/models/policies.ts
jeffrimko: Updated in new commit.

src/sdk/resources/policyActions.ts
jeffrimko: Updated in new commit.

jeffrimko:
@mmitoff @jtirjan Can you take another quick look at this PR? Think all your prior comments have been addressed.

src/sdk/models/policies.ts
```diff
@@ -86,14 +86,18 @@ export interface PolicyTrackingGroupRule extends IEntityIdentifier<string> {
 export interface PolicyActionDescriptor extends IEntityIdentifier<string> {
 	name?: string;
 	description?: string;
+	destructive?: boolean;
```
jtirjan, line 89: If we put these properties here, then `PolicyTrackingGroupActionConfiguration` can't extend this interface, because it doesn't have any of these.
Might be best to do:
```ts
interface PolicyActionBase extends IEntityIdentifier<string> {
	name?: string;
	description?: string;
}

export interface PolicyActionDescriptor extends PolicyActionBase {
	destructive?: boolean;
	exclusive?: boolean;
	requires_existing?: boolean;
	removes_item?: boolean;
	prompt?: any;
}

export interface PolicyTrackingGroupActionConfiguration extends PolicyActionBase {
	[name: string]: any;
}
```

jtirjan (changes requested):

src/sdk/models/policies.ts
jtirjan: Works for me

src/sdk/models/policies.ts
jeffrimko: Updated in new commit 21eb0e4.

src/sdk/resources/policyActionTypes.ts
```diff
@@ -0,0 +1,9 @@
+import { IHttpClient } from '../http';
+import { PagedResource } from './resource';
+import { PolicyActionDescriptor } from '../models';
+
+export class PolicyActionTypesResource extends PagedResource<PolicyActionDescriptor> {
```
jtirjan, line 5: Sorry, just noticed this. I think we just want this to extend `Resource` (get & list).

```suggestion
import { Resource } from './resource';
import { PolicyActionDescriptor } from '../models';

export class PolicyActionTypesResource extends Resource<PolicyActionDescriptor> {
```

jtirjan (changes requested):

jtirjan (approved):

PR #193, opened 2021-07-12
ACT-1486 | UX | Policy Editor - Manual intervention on actions

jeffrimko:
Added PolicyActionTypeId enum.

### DIFF
--- a/src/sdk/models/policies.ts
+++ b/src/sdk/models/policies.ts
@@ -95,7 +95,7 @@ export interface PolicyTrackingGroupRule extends IEntityIdentifier<string> {
 	criteria?: MetadataCalculatedFilter;
 }
 
-interface PolicyActionBase extends IEntityIdentifier<string> {
+interface PolicyActionBase extends IEntityIdentifier<PolicyActionTypeID | string> {
 	name?: string;
 	description?: string;
 }
--- a/src/sdk/models/policies.ts
+++ b/src/sdk/models/policies.ts
@@ -25,6 +25,14 @@ export enum PolicyStatus {
 	NoIssues = 'no-issues'
 }
 
+export enum PolicyActionTypeID {
+	Approval = 'approval',
+	Delay = 'delay',
+	Delete = 'delete',
+	Move = 'move',
+	Remediate = 'remediate'
+}
+
 export interface Policy extends IEntityIdentifier<string>, IAuditedEntity {
 	name?: string;
 	description?: string;


src/sdk/models/policies.ts
```diff
@@ -25,6 +25,14 @@ export enum PolicyStatus {
 	NoIssues = 'no-issues'
 }
 
+export enum PolicyActionTypeId {
```
smbecker, line 28:
```suggestion
export enum PolicyActionTypeID {
```

smbecker (changes requested):

src/sdk/models/policies.ts
```diff
@@ -83,7 +91,7 @@ export interface PolicyTrackingGroupRule extends IEntityIdentifier<string> {
 	criteria?: MetadataCalculatedFilter;
 }
 
-interface PolicyActionBase extends IEntityIdentifier<string> {
+interface PolicyActionBase extends IEntityIdentifier<PolicyActionTypeID> {
```
gotvino, line 94: Shouldn't this still allow custom ids? @smbecker thoughts?
```suggestion
interface PolicyActionBase extends IEntityIdentifier<PolicyActionTypeID | string> {
```

gotvino (dismissed):
Rebase and [this](https://github.com/portalarchitects/skysync-cli/blob/480a678c398c1ea8f3b9df2e1ca78e4d1054a8bd/src/sdk/models/policies.ts#L69) may need to be:
```
	[action: PolicyActionTypeID | string]: any;
```

jeffrimko:
For now, only going to add the new enum type and not use it for any IEntityIdentifier definitions or similar.

gotvino:
> For now, only going to add the new enum type and not use it for any IEntityIdentifier definitions or similar.

The suggested change would have worked fine, no?

jeffrimko:
> > For now, only going to add the new enum type and not use it for any IEntityIdentifier definitions or similar.
> 
> The suggested change would have worked fine, no?

Yes and we can change it back if desired. I'm not familiar enough with how the IDs will be used going forward so just keeping it a string seems like the safest bet for now.

src/sdk/models/policies.ts
jtirjan: I agree with @gotvino on this one, because it self-documents the relationship, while still allowing for a `string` fallback.

jeffrimko: Okay, it has been added back.

jtirjan (approved):

PR #198, opened 2021-07-29
SF-7962 | Added support for accessing Fiddler API calls

Alex-Murtha:


src/sdk/models/diagnostics.ts
```diff
@@ -3,3 +3,8 @@ export interface DiagnosticMetric  {
 	value: string;
 	type: string;
 }
+
+export interface DiagnosticFiddlerStatus {
```
MicheleBuselli, line 7: Looks like there's a similar interface [here](https://github.com/portalarchitects/skysync-cli/blob/5fcad3a564e291787e5a7dfd91de47169292d951/src/commands/diagnostics/fiddler/util.ts#L20)

MicheleBuselli (changes requested):

src/sdk/models/diagnostics.ts
Alex-Murtha: Removed the duplicate interface and referenced the exported SDK interface

MicheleBuselli (approved):

PR #202, opened 2021-08-04
ACT-1098 | statistics summary resource

bentatum:
don't merge, not sure if this is correct - yet.

### DIFF
--- a/src/sdk/models/policyStatistics.ts
+++ b/src/sdk/models/policyStatistics.ts
@@ -40,7 +40,7 @@ export interface PolicyStatistic extends LabelledStatistic, StatisticValue {
 	remediated?: StatisticValue;
 }
 
-export type PolicyStatisticsTimeline = { 
+export type PolicyStatisticsTimeline = {
 	timestamp?: number;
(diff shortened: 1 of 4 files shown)


src/sdk/models/policyStatistics.ts
```diff
@@ -61,3 +61,14 @@ export interface PolicyStatistics extends IEntityIdentifier<string> {
...
+
+export interface PolicySummaryByRisk {
+	bytes: number;
+	files: number;
+	name: string;
+	order: number;
+}
+
+export interface PoliciesStatisticsSummary {
+	policies_by_risk?: PolicySummaryByRisk[];
```
avolinsky, line 73: I think this can be
```suggestion
	policies_by_risk?: PolicyStatistic[];
```

But you would need to add `policy?: Policy` prop [here](https://github.com/portalarchitects/skysync-cli/blob/ff24083784964a3f2640bdc934167a0207055530/src/sdk/models/policyStatistics.ts#L38)

avolinsky (changes requested):

avolinsky (approved):

PR #210, opened 2021-08-24
ACT-1873 | Add GET Configuration endpoint to Extensions resource

mmitoff:
Add GET Configuration endpoint to Extensions resource

### DIFF
--- a/src/sdk/resources/extensions.ts
+++ b/src/sdk/resources/extensions.ts
@@ -36,9 +36,8 @@ export class ExtensionsResource extends Resource<Extension> implements IDownload
 		return this.getSingle(result);
 	}
 
-	async updateConfiguration(id: string, values: any, params?: any, token?: CancellationToken): Promise<any> {
-		const response = await this.httpClient.patch(`${this.resourcePath}/${id}/configure`, values, this.mergeDefaultParams(params), token);
-		return getTypedResponse<Connection>(response);
+	async updateConfiguration(id: string, values: any, params?: any, token?: CancellationToken): Promise<{}> {
+		return await this.httpClient.patch(`${this.resourcePath}/${id}/configure`, values, this.mergeDefaultParams(params), token);
 	}
 
 	getConfiguration(id: string, params?: any, token?: CancellationToken): Promise<{}> {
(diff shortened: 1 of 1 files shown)


src/sdk/resources/extensions.ts
```diff
@@ -36,11 +36,15 @@ export class ExtensionsResource extends Resource<Extension> implements IDownload
...
 
-	async configure(id: string, values: any, params?: any, token?: CancellationToken): Promise<Extension> {
+	async updateConfiguration(id: string, values: any, params?: any, token?: CancellationToken): Promise<Extension> {
 		const response = await this.httpClient.patch(`${this.resourcePath}/${id}/configure`, values, this.mergeDefaultParams(params), token);
 		return getTypedResponse<Connection>(response);
 	}
 
+	async getConfiguration(id: string, params?: any, token?: CancellationToken): Promise<{}> {
+		return await this.httpClient.get(`${this.resourcePath}/${id}/configure`, this.mergeDefaultParams(params), token);
+	}
```
jtirjan, line 46:
```suggestion
	getConfiguration(id: string, params?: any, token?: CancellationToken): Promise<{}> {
		return this.httpClient.get(`${this.resourcePath}/${id}/configure`, this.mergeDefaultParams(params), token);
	}
```

src/sdk/resources/extensions.ts
```diff
@@ -36,11 +36,15 @@ export class ExtensionsResource extends Resource<Extension> implements IDownload
 		return this.getSingle(result);
 	}
 
-	async configure(id: string, values: any, params?: any, token?: CancellationToken): Promise<Extension> {
+	async updateConfiguration(id: string, values: any, params?: any, token?: CancellationToken): Promise<Extension> {
```
jtirjan, line 39: This doesn't actually match up with `Extension`. It has `config_attributes` instead of `configuration`.
Not sure why it's named differently here.
```suggestion
	async updateConfiguration(id: string, values: any, params?: any, token?: CancellationToken): Promise<any> {
```

jtirjan (changes requested):
A couple super minor requests

src/sdk/resources/extensions.ts
mmitoff: Good call

jtirjan (approved):

PR #215, opened 2021-08-30
SF-7795 | Added new CLI commands for managing parallel write configuration

Alex-Murtha:
Back end changes in this [PR](https://github.com/portalarchitects/skysync-vnext/pull/3484)

### DIFF
--- a/src/commands/performance/parallel_writes/update.ts
+++ b/src/commands/performance/parallel_writes/update.ts
@@ -35,7 +35,7 @@ export = {
 			};
 
 			if (!performance.parallel_writes.requested && !performance.parallel_writes.max) {
-				throw exception()
+				throw exception();
 			}
 
 			writeConfiguration(await client.performance.post(performance), output);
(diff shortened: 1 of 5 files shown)


src/commands/performance/parallel_writes/update.ts
```diff
@@ -0,0 +1,33 @@
+import { runCommand } from '../../../util/command';
+import { writeConfiguration } from './util';
+import { Performance } from './../../../sdk';
+
+export = {
+	command: 'update',
+	description: 'Display current performance configuration',
```
MicheleBuselli, line 7: 'Update' instead of 'Display'?

MicheleBuselli (changes requested):
Not sure how to see the options for the update command. When I type 'skysync.js performance parallel_writes update' it returns the table, but shouldn't I see what the options are (requested and  max)?

src/commands/performance/parallel_writes/update.ts
Alex-Murtha: Done

MicheleBuselli (approved):

PR #216, opened 2021-08-30
SF-9179 | added POST functionality to the get connection items call to handle long URLs

alanconnor:
when passing the next page token for listing items from a connection the "Load More" link sent the data in the query sting as a GET.  DropBox had next page tokens that would make the URL too long and a http status of 414 was returned.  The code will now look at the length of the URL and perform a POST of the token.


Alex-Murtha (approved):
Tested changes with this change through vnext PR

PR #218, opened 2021-08-31
sf-9045 | Expose get/set of logging status as commands

MicheleBuselli:


src/commands/diagnostics/logging/update.ts
```diff
@@ -0,0 +1,38 @@
...
+export = {
+	command: 'update',
+	description: 'Update logging configuration',
+	builder: yargs => {
+		yargs.options({
+			'level': {
+				alias: 'l',
+				desc: 'Set the level of detail for application logs',
+				type: 'string'
+			},
```
Alex-Murtha, line 14: We should take advantage of yargs `choices` feature for log level: https://yargs.js.org/docs/#api-reference-choiceskey-choices

Alex-Murtha (changes requested):

src/commands/diagnostics/logging/util.ts
```diff
@@ -0,0 +1,33 @@
+import { DiagnosticLoggingStatus, LoggingLevel } from '../../../sdk';
+import { OutputFormatter } from '../../../util/formatter';
+
+export const getLevels = (enumObject: typeof LoggingLevel) : string[] => {
+	let levels = [];
+	for (let key in enumObject) {
+		if ({}.hasOwnProperty.call(enumObject, key)) {
```
MicheleBuselli, line 7: This is weird but it gets rid of an eslint warning

Alex-Murtha: Try using this:
```
export const getLogLevels = () : string[] => {
  return Object.keys(LoggingLevel).map(key => LoggingLevel[key]); // possibly use Object.values(LoggingLevel)?
}
```
https://stackoverflow.com/questions/39372804/how-can-i-loop-through-enum-values-for-display-in-radio-buttons

Alex-Murtha: Might be worth moving this to models as well so the UI can use this function and not have duplicate logic

Alex-Murtha (approved):

PR #219, opened 2021-09-02
ACT-1109 | Add formatCurrency to DataFormatter

mmitoff:
Add formatCurrency to DataFormatter

### DIFF
--- a/src/sdk/formatting/formatCurrency.ts
+++ b/src/sdk/formatting/formatCurrency.ts
@@ -1,0 +1,7 @@
+import { formatNumber } from './formatNumber';
+
+export const formatCurrency = (number: number, locale: string = 'en', currency: string = 'USD'): string =>
+	formatNumber(number, null, new Intl.NumberFormat(locale, {
+		style: 'currency',
+		currency
+	}));
--- a/src/sdk/formatting/formatCurrency.ts
+++ b/src/sdk/formatting/formatCurrency.ts
@@ -1,7 +1,9 @@
 import { formatNumber } from './formatNumber';
 
-export const formatCurrency = (number: number, locale: string = 'en', currency: string = 'USD'): string =>
+export const formatCurrency = (number: number, locale: string = 'en', currency: string = 'USD', wholeUnitsOnly: boolean = false): string =>
 	formatNumber(number, null, new Intl.NumberFormat(locale, {
 		style: 'currency',
-		currency
+		currency,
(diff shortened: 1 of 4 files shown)


src/sdk/formatting/formatCurrency.ts
```diff
@@ -0,0 +1,7 @@
+import { formatNumber } from './formatNumber';
+
+export const formatCurrency = (number: number, locale: string = 'en', currency: string = 'USD'): string =>
+	formatNumber(number, null, new Intl.NumberFormat(locale, {
+		style: 'currency',
+		currency
+	}));
```
jtirjan, line 7: Can you also add support for `wholeUnitsOnly` (or a better name if you can think of one)
```ts
export const formatCurrency = (number: number, locale: string = 'en', currency: string = 'USD', wholeUnitsOnly: boolean = false): string =>
	formatNumber(number, null, new Intl.NumberFormat(locale, {
		style: 'currency',
		currency,
		maximumFractionDigits: wholeUnitsOnly ? 0 : undefined
	}));
```

jtirjan (approved):

src/sdk/formatting/formatCurrency.ts
```diff
@@ -0,0 +1,9 @@
+import { formatNumber } from './formatNumber';
+
+export const formatCurrency = (number: number, locale: string = 'en', currency: string = 'USD', wholeUnitsOnly: boolean = false): string =>
```
smbecker, line 3: We should follow convention and add `CurrencyFormatOptions = { locale?: string, currency?: string }`

src/sdk/formatting/formatNumber.ts
```diff
@@ -12,7 +12,7 @@ export type NumberFormatOptions = {
 	unit?: string;
 };
 
-export const formatNumber = (number: number, options?: NumberFormatOptions): string => {
+export const formatNumber = (number: number, options?: NumberFormatOptions, formatter: Intl.NumberFormat = numberFormatter): string => {
```
smbecker, line 15: I don't think that `formatter` should be publicly exposed here. Since `options` is always being passed in from `formatCurrency` as `null`, there really is no value in leveraging this method from `formatCurrency` and it could be simplified to
```js
if (!Boolean(number) || isNaN(number)) {
	number = 0;
}
return formatter.format(number);
```

mmitoff: @smbecker Made these changes [here](https://github.com/portalarchitects/skysync-cli/pull/220)

PR #243, opened 2022-01-14
ACT-2883 | Updated dependecies

avolinsky:


src/sdk/resources/personalDrive.ts
```diff
@@ -1,6 +1,7 @@
 import { IHttpClient } from '../http';
 import { PagedResource } from './resource';
 import { PersonalDriveMatchesResource } from './personalDriveMatches';
+import { Account } from '..';
```
gotvino, line 4: Why was this needed?

avolinsky: `Account` was part of global API until recent TypeScript update..

gotvino (approved):

PR #260, opened 2022-11-17
CLI: Error being returned when trying to run commands

DinuAlin:
Starting with node version [18.0.0](https://github.com/nodejs/node/releases/tag/v18.0.0) an experimental fetch API was enabled by default which was breaking our commands. In order to fix this we have to add the `--no-experimental-fetch` flag.

### DIFF
--- a/src/sdk/http/http-client-factory.ts
+++ b/src/sdk/http/http-client-factory.ts
@@ -5,22 +5,19 @@ declare const window: any;
 declare const WorkerGlobalScope: any;
 declare const global: any;
 
-function getFetch(): {fetch: FetchApi; formDataType: any; usingNodeFetch: boolean} {
+function getFetch(): {fetch: FetchApi; formDataType: any } {
 	return (function(g: any) {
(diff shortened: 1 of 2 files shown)


src/sdk/http/http-client-factory.ts
```diff
@@ -5,22 +5,19 @@ declare const window: any;
 declare const WorkerGlobalScope: any;
 declare const global: any;
 
-function getFetch(): {fetch: FetchApi; formDataType: any; usingNodeFetch: boolean} {
+function getFetch(): {fetch: FetchApi; formDataType: any } {
 	return (function(g: any) {
 		let f = g?.fetch;
 		let formDataType = g?.FormData;
-		let usingNodeFetch = false;
```
smbecker, line 12: Rather than removing, we should perhaps use `!Boolean(g?.window)`.

smbecker (changes requested):

smbecker (approved):

